PR #10, opened 2021-04-26
feat: Adding storefront/product

kysely:
Adding `storefront/product` capability with 2 providers:

- Stripe (uses recursion for pagination of subresources)
- Shopify (uses GraphQL)

This will be more useful once combined with other Storefront (and/or eCommerce) capabilities to build vendor-agnostic stores (headless commerce, etc).

Screenshot 2021-04-26 at 13 41 57


superface/super.json
```diff
@@ -108,6 +108,24 @@
         }
       }
     },
+    "storefront/product": {
+      "file": "./grid/storefront/product/profile.supr",
```
freaz, line 112: I know it is caused, because this PR was opened already long ago. But the paths should point to files in `capabilities` and `providers` folders.

Jakub-Vacek:
Closed after talking with Radek

PR #21, opened 2021-05-07
Speech synthesis

janhalama:
This PR adds `speech/synthesis` capability with Google apis and IBM cloud providers.

This PR also renames `GOOGLE_MAPS_API_KEY` to `GOOGLE_APIS_API_KEY` so that it can be shared by Google apis providers.

Representation of `audioContent` in snapshots has thousands of lines of code. It is generated file, I am not sure if this can be optimised somehow.


zdne (approved):
Does it work? 😎 I am mind blown! Great job @janhalama ! 👏

PR #66, opened 2021-09-14
feat: Add navigation/nearby-poi capability, overpass-de provider

TheEdward162:
## Description

This capability can search for nearby (latitude+longitude+radius) points of interest with overpass (open street maps) and google maps providers. The profile currently only supports filtering by RESTAURANT, CAFE and BAR categories, but can later be extended to support other categories.


freaz (commented):
Profile and maps look great. 

One thing I would do to avoid troubles in future is to introduce new provider for google apis. google-apis name is generic, but allows to use only maps.googleapis.com. So we will need to do something about that provider.

Also could you please use testing-lib ([example](https://github.com/superfaceai/station/blob/main/capabilities/communication/send-email/maps/send-email.ts)) for tests? And remove generated types as they won't be needed while using testing-lib?

capabilities/navigation/nearby-poi/maps/google-apis-maps.suma
```diff
@@ -0,0 +1,108 @@
...
+        )
+      }
+
+      map result points
+    }
+  }
+}
+
+operation BuildCategoryQueryType {
+  return if (args.categories.length !== 1) undefined
```
martinalbert, line 39: This will return undefined when there will be more than one categories, is it intentional?

```suggestion
  return if (args.categories.length === 0) undefined
```

capabilities/navigation/nearby-poi/maps/google-apis-maps.suma
```diff
@@ -0,0 +1,108 @@
+profile = "navigation/nearby-poi@1.0"
+provider = "google-apis-maps"
+
+map NearbyPoi {
```
martinalbert, line 4: Is this intentional? Shouldn't this be `NearbyPoint`?

martinalbert: Same goes to profile and usecases. 🤷🏻‍♂️

capabilities/navigation/nearby-poi/profile.supr
```diff
@@ -0,0 +1,65 @@
+"""
+Find nearby points of interest.
+
+Find points of interest near the given location using map services.
+"""
+name = "navigation/nearby-poi"
```
martinalbert, line 6: Shouldn't name be `navigation/nearby-point`?

capabilities/navigation/nearby-poi/profile.supr
```diff
@@ -0,0 +1,65 @@
...
+
+Find points of interest near the given location.
+"""
+usecase NearbyPoi {
+	input {
+		"""
+		center
+		Center of the search
+		"""
+		center! Coordinates!
```
martinalbert, line 20: Space indentation seems weird in this file.

capabilities/navigation/nearby-poi/profile.supr
```diff
@@ -0,0 +1,65 @@
...
+model Coordinates {
+	latitude! number!
+	longitude! number!
+}
+
+model InterestCategory enum {
+	RESTAURANT,
+	CAFE,
+	BAR
+	// TODO: More categories
```
martinalbert, line 64: Can you please add some more categories from [here](https://developers.google.com/maps/documentation/places/web-service/supported_types) before we publish this? Probably most generic ones to be able to work with overpass as well.

martinalbert (changes requested):

capabilities/navigation/nearby-poi/maps/google-apis-maps.suma
Jakub-Vacek: POI => Point Of Interest 🤷

martinalbert: Oh, ok 😄

martinalbert:
I was thinking whether this should have scope `address` rather than `navigation`. @Jakub-Vacek what do you think?

capabilities/navigation/nearby-poi/maps/google-apis-maps.suma
TheEdward162: Sorry, I forgot to comment this. It returns `undefined` for anything but one category because google apis don't allow specifying multiple categories.

When the user requests one category, the map should propagate that to the API. When the user requests zero, two or more categories, we should not pass any category filter to the API, otherwise they would only get the first category they wanted and no other.

martinalbert: Oh right, categories are passed as array, so they're formated like `param,param`. Would formatting categories like `param|param` work? Based on their [docs](https://developers.google.com/maps/documentation/places/web-service/search-nearby#type) it takes the first specified, but still better than no category. 🤷🏻‍♂️

TheEdward162: I don't think it's better, because then it will never return any category other than the first, which directly contradicts the idea of requesting multiple categories.

I choose this because it will return all POIs in the vicinity and then there is a filter [here](https://github.com/superfaceai/station/pull/66/files/f40beb2a0762c2ebd0adec3c5e93299f5f490a00#diff-657a37b57be357898afc628af21da103e7e79f4a2c1b053cabcc0b1ae4ee17e7R28) that makes sure only the specified categories are actually returned by the map.

The big downside of this is that the accuracy will be severely limited since the API returns many other categories so the selected ones will be spread out.

My only other idea was to run the request for each category, but that might be an unexpected cost to the user because it will fire off several requests, each of which will be billed by Google.

martinalbert: Oh nice, I didn't understood that filter. This is definitely better.

martinalbert: @Jakub-Vacek can you please revert suggested change?

martinalbert (approved):

Jakub-Vacek:
@TheEdward162 @martinalbert @freaz @janhalama This should be ready - can I merge it? 🤷🙂

freaz:
:shipit:

PR #70, opened 2021-09-16
Feat/add openweathermap provider

kravchenkoa:
## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. For updating Oclif commands documentation use [oclif-dev](https://github.com/oclif/dev-cli#oclif-dev-readme).


providers/openweathermap.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "openweathermap",
+  "services": [
+    {
+      "id": "default",
+      "baseUrl": "https://api.openweathermap.org"
```
jnv, line 6: I think it's preferable to include version prefixes into provider's `baseUrl` (or so it's my assumption looking at existing providers), so:

```suggestion
      "baseUrl": "https://api.openweathermap.org/data/2.5"
```
(and update the map's path accordingly)

But maybe the practice has changed since this kinda defies the idea of immutability of providers?

kravchenkoa: That depends on how they version the API - if any update increments the version for all, or if different APIs can be on different version (e.g. /weather and /forecast). I think not including it in the base url is safer.

zdne: I think @kravchenkoa is correct with his argument here! +1

jnv: And Ondra already added that to the conventions. Very well!

jnv (approved):

zdne (approved):
👍🏻

PR #86, opened 2021-10-05
feat: Add language translation capability

martinalbert:
## Description
Adds capability for text translation.

## Types of changes

## Checklist:


.eslintrc.js
```diff
@@ -5,12 +5,8 @@ module.exports = {
     tsconfigRootDir: __dirname,
     project: ['./tsconfig.json'],
   },
-  ignorePatterns: ["superface/**/*.js", "superface/types", "*.config.js"],
```
freaz, line 8: Why the formatting is changes? It is not happening with `yarn format:fix` and `yarn lint:fix` on shared `chore/update-sf-dependencies`.

freaz (commented):
Last three commits would be better separated from this PR.

Especially formatting, which I still don't know from where is it coming.
Also updating testing lib, because this way we will need to get this to shared branch with bunch of other questionable changes.

.eslintrc.js
martinalbert: I'll revert those 3 commits and make it work without these weird changes. 👍🏻

PR #158, opened 2021-11-15
Add `analytics/track-event`

freaz:


superface/super.json
```diff
@@ -325,6 +325,20 @@
...
+      "file": "../capabilities/analytics/track-event/profile.supr",
+      "providers": {
+        "mixpanel": {
+          "file": "../capabilities/analytics/track-event/maps/mixpanel.suma"
+        },
+        "mock": {
+          "file": "../capabilities/analytics/track-event/maps/mock.suma"
+        }
+      },
+      "priority": [
```
Jakub-Vacek, line 339: priority is not needed :)

freaz: Added automatically by CLI. I didn't manually modified it.

capabilities/analytics/track-event/maps/mixpanel.suma
```diff
@@ -0,0 +1,39 @@
+// API Reference: https://developer.mixpanel.com/reference/events
+
+profile = "analytics/track-event@1.0"
+provider = "mixpanel"
+
+map TrackEvent {
```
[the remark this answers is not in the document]
freaz: It is, but because Mixpanel is adding token to different places in body and they say project token isn't secret, I decided to use integration parameter and add it where it is needed.

superface/super.json
Jakub-Vacek: Yep, cli doesn't know we don't need it in station. I usually delete priority from added stuff in station super.json 🤷

freaz: Removed

janhalama (approved):
Looks good 👍

PR #163, opened 2021-11-22
Implement mixpanel for `crm/contacts`

freaz:
## Description

It is a bit controversial, because Mixpanel isn't actually CRM and doesn't allow creating new contacts (users).
You as developer need to send unique id of the user, and if it doesn't exists it will be created.

Also Mixpanel doesn't does not expose users filtering in public API.

It share provider from https://github.com/superfaceai/station/pull/158

## Types of changes

## Checklist:


superface/super.json
```diff
@@ -297,12 +297,21 @@
         },
         "hubspot": {
           "file": "../capabilities/crm/contacts/maps/hubspot.suma"
+        },
+        "mixpanel": {
+          "file": "../capabilities/crm/contacts/maps/mixpanel.suma"
         }
-      }
+      },
+      "priority": [
```
Jakub-Vacek, line 305: Maybe remove priority 🤷

Jakub-Vacek (approved):

superface/super.json
freaz: Shouldn’t be priority added by CLI only of there is more than one provoder configured? 

What it changes if it is set?

How sdk behaves when concrete provoder is passed in runtime? Does it failover anyway?

Jakub-Vacek: I will look into CLI behavior.

It changes nothing, it's just dead code. You can keep it, my OCD kicked in 😅

freaz: Ok, I am curious as it is what users will have in super.json as I am not expecting, they will be removing it.

freaz: Hmm when I went to fix it, I noticed there is more providers. When I specify priority and omit some providers does it meant fallback to them won't happen?

For sure this is under documented.

PR #165, opened 2021-11-22
Remove generated types for OneSDK

freaz:
## Description

- Removes generated types
- Fixes mock tests using generated SDK

## Motivation and Context

Generated types are experimental and it is not documented. Since everywhere except mock providers is used testing library, I would hide this until we decided it is ready.

If needed, we can add step generating types and saving them as artefacts to see how generation works. 

## Types of changes

## Checklist:


Jakub-Vacek (approved):
Code is OK, but we should probably start thinking about making typed interface usable (tested, documented). Currently it fells like we have something which is 80% ready but hidden. What do you think @lukas-valenta ?

lukas-valenta:
i think we have other priorities now and it would be better to stick to javascript to avoid bugs and confusion until we're stable enough

freaz:
Agree with Luky. Internally it is still used, so some usage and testing it has, but I'd say it needs proper plan for release.

PR #170, opened 2021-12-02
feat: Crisp integration for `crm/contacts`

kysely:
## Description

Adding [Crisp chat integration](https://docs.crisp.chat/guides/rest-api/quickstart/) to [crm/contacts](https://superface.ai/crm/contacts).

### Few concerns

Crisp uses a separate subresource for contact's custom properties (they call it “custom data”). This brings several challenges when mutating the data in a single use case.

#### `Create` use case
- first creates the general contact
  - if that fails (usually the contact already exists), all is good, nothing in Crisp was changed, can be retried
- then updates the contact's custom data subresource
  - that fails, we have inconsistent data in Crisp. The _new_ contact was created, but custom data was not set

I chose to fail the whole `Create` use case when custom data patch fails with a custom error describing the issue. No one will actually read the map and implement the special case, but:
- they probably will retry the failed `Create`
- since the contact will exist, it will return 404 (Already exists)
- that should be switched to `Update` (generally)

Alternatively, we could simply _not support_ this use case (like [Mixpanel](https://superface.ai/raw/crm/contacts.mixpanel@1.0.1.suma) integration) and quietly create the non-existent resource in `Update`.

#### `Update` use case
- first updates the general contact
  - if that fails (404 or smth), all is good, nothing was changed, can be retried
- then updates the contact's custom data subresource
  - that fails, we have inconsistent data in Crisp. The contact was updated, but custom data was not set

This is much easier since `Update` is much more idempotent-ish. We simply fail the whole use case call and it can be safely retried, yielding the desired end result.

## Types of changes

## Checklist:


grid/crm/contacts/maps/crisp.suma
```diff
@@ -0,0 +1,286 @@
...
+
+      map result {
+        id = body.data.people_id
+      }
+    }
+
+    response 409 {
+      map error {
+        title = "Contact already exists"
+        detail = (`Contact '${input.email}' already exists. If you with to update it, use 'Update' use case`)
```
TheEdward162, line 45:
```suggestion
        detail = (`Contact '${input.email}' already exists. If you wish to update it, use 'Update' use case`)
```

grid/crm/contacts/maps/crisp.suma
```diff
@@ -0,0 +1,286 @@
...
+    model: "people"
+  }]
+
+  search_filter_query_param = encodeURIComponent(JSON.stringify(search_filter))
+
+  http GET "/v1/website/{parameters.WEBSITE_ID}/people/profiles?search_filter={search_filter_query_param}" {
+    security "plugin_tier"
+
+    request "application/json" {
+      headers { "X-Crisp-Tier" = "plugin" }
```
TheEdward162, line 121: Is there a problem with using `query { search_filter = search_filter }` here? (note it has to go before the `headers { ... }` slot, as the ordering is fixed right now).

grid/crm/contacts/maps/crisp.suma
```diff
@@ -0,0 +1,286 @@
...
+    title = "Not allowed"
+  }
+
+  return if (statusCode === 404) {
+    title = "Not found"
+    detail = "Contact or website was not found or Crisp plugin is not subscribed. Make sure the Crisp website exists and uses your plugin"
+  }
+
+  return if (statusCode === 423 || statusCode === 429) {
+    title = "Quota limit exceeded!"
```
TheEdward162, line 274: I am a bit put off by the texts here. Sometimes there are periods at the end, sometimes there aren't and the one title has an exclamation mark! 😱

grid/crm/contacts/maps/crisp.suma
kysely: Guess I had a blackout.... `query` slot works well :)

kysely:
@TheEdward162 Thanks! I worked in all the suggestions :)

grid/crm/contacts/maps/crisp.suma
```diff
@@ -0,0 +1,286 @@
...
+// Utilities
+operation BuildCrispProfile {
+  input = args.input
+
+  fullName = [input.firstName || "", input.lastName || ""].join(' ').trim()
+
+  return { 
+    email: input.email,
+
+    person: fullName.length || input.phone || input.country ? {
```
Jakub-Vacek, line 193: This construction is pretty confusing 🤨

kysely: Hmm, what exactly about it? 🤔  The fact that it's inlined, or the actual use of ternary for this?

The `person` object optional, but if it's present, some fields inside are required, that's why I need to always check whether to even construct it.

Alternatively I had bunch of `set if () {...}` for each field but it was one screen long, and not much more readable.

What do you have in mind for this?

Jakub-Vacek (approved):
This is pretty complicated map, I like how you reused some operations 👍

grid/crm/contacts/maps/crisp.suma
Jakub-Vacek: Well, the combination on inline and ternary is imo hard to read. Bunch of set if will be probably too long. 
It would be nice to do something like:
`set if(fullName.length || input.phone || input.country) {
  person = {...}
}`

Not sure that will work :/
Anyway it's not a blocker 🙂

kysely: Those `set if() { person ... }` statements have problem that you cannot reference `person` if the condition evaluates to false. I need it to be undefined.

I was able to simplify it the following way, is it more readable?

```graphql
operation BuildCrispProfile {
  input = args.input

  fullName = [input.firstName || "", input.lastName || ""].join(' ').trim() || undefined

  return { 
    email: input.email,

    person: (fullName || input.phone || input.country) && {
      nickname: fullName,
      phone: input.phone,
      geolocation: { country: input.country }
    },

    company: input.company && {
      name: input.company
    }
  }
}
```

Jakub-Vacek: Yep, I like that 👍👍

PR #176, opened 2022-01-10
feat(communication/send-email): Add attachments

tjerabek:
## Description

This PR adds sending attachments into `send-email` use case. Attachments are an array of objects. Content of an attachment is `base64` encoded file.

## Motivation and Context

This solves need for sending attachments using email providers.

## Types of changes

## Checklist:


grid/communication/send-email/maps/postmark.suma
```diff
@@ -4,6 +4,17 @@ profile = "communication/send-email@2.0"
 provider = "postmark"
 
 map SendEmail {
+  Attachments = null
```
freaz, line 7: I don't see Attachments used anywhere. How it gets to request?

grid/communication/send-email/maps/sendgrid.suma
```diff
@@ -21,6 +22,16 @@ map SendEmail {
     content = content.concat([{ type: 'text/html', value: input.html }])
   }
 
+  set if (input.attachments) {
+    attachments = input.attachments.map(attachment => {
+      return {
```
freaz, line 27: Map seem unnecessary, as it maps 1:1 input to what needs to be in request.

jnv: I am not 100% sure how profile-validation works on input, but I _think_ it let's you keep additional object properties, which would be passed to the provider without explicit mapping. Also this marks a clear extension point if the profile changes, so personally I am in favor of having explicit mappings like this.

freaz (approved):
I am not on same page with @jnv about unnecessary mappings, but as you two agree 🤷 

Otherwise looks good. 👍

jnv:
Note that the implementation for Mailgun is incorrect, it's supposed to send attachments as `multipart/form-data` but that's not the case. I would prefer to fix that separately to _finally merge this_.

PR #185, opened 2022-01-26
Add `social-media/posts-lookup` profile with Twitter map

janhalama:
## Description

This PR adds `social-media/posts-lookup` profile with `FindByHashtag` and `FindByMention` use-cases. PR adds also maps for  `mock` and `twitter` providers.

## Types of changes

## Checklist:


grid/social-media/content-search/maps/twitter.suma
```diff
@@ -0,0 +1,168 @@
...
+
+profile = "social-media/content-search@1.0"
+provider = "twitter"
+
+map SearchByHashtag {
+
+  http GET "/2/tweets/search/recent" {
+
+    
+
```
jnv, line 11: Formatting? :thinking:

jnv (commented):
I have the following concerns about this combination of use-cases:

- the naming feels somewhat misleading, I'd expect a fulltext search under this profile
- not all SM allow searching both with hashtags and mentions, so this might target only Twitter and Twitter-like media (Mastodon comes to mind) - and on the other hand, not all media support searching by fulltext
- I think we don't have any use-case for fetching the profile ID from username in case of Twitter; IMO the `SearchByMentions` could accept `username` as an alternative to `profileId`? (can be extended later)

So while I hate to suggest it, I'd consider renaming the profile and use-cases to better align expectations.
- I don't mind keeping the use-cases under the single profile as long as they remain separate from full-text search
- rename `content-search` to `content-lookup`, or `posts-lookup`
- use cases from `SearchBy*` to `FindByHashtag`, `FindByMention` (mind the singular here)

"Lookup" instead of search implies that you have better idea what you are looking for, while search is rather more general. But I welcome any other suggestions.

janhalama:
>"Lookup" instead of search implies that you have better idea what you are looking for, while search is rather more general. But I welcome any other suggestions.

I understand your concerns, Lookup sounds much more appropriate in this case.

janhalama:
> I think we don't have any use-case for fetching the profile ID from username in case of Twitter; IMO the SearchByMentions could accept username as an alternative to profileId? (can be extended later)

Right we do not have it. `username` seems to be appropriate naming as even Facebook requires to create `username` for Facebook page to be able to mention it.

grid/social-media/content-search/maps/twitter.suma
janhalama: fixed

janhalama:
@jnv I changed the name of profile and use-cases as you proposed. 🙏

jnv (approved):
LGTM, thanks!

grid/social-media/posts-lookup/maps/twitter.suma
```diff
@@ -0,0 +1,166 @@
...
+    
+    request {
+      query {
+        query = '#' + input.hashtag
+        'tweet.fields' = 'id,text,created_at,attachments'
+        expansions = 'attachments.media_keys'
+        'media.fields' = 'duration_ms,height,media_key,preview_image_url,type,url,width,alt_text'
+        start_time = input.afterDate
+        pagination_token = input.page
+        max_results = 100
```
martinalbert, line 18: I'm not sure whether it matters, but snapshots are quite long for this profile, could we use input parameter limit to have less records to compare during testing?

martinalbert (approved):

grid/social-media/posts-lookup/maps/twitter.suma
janhalama: 👍 I will add the page limit parameter in separate PR.

PR #187, opened 2022-01-27
feat: Add `chat/messages` profile and slack map

martinalbert:
## Description
Adds profile for getting past messages in slack.

## Types of changes

## Checklist:


janhalama:
@martinalbert is this ready for review? There are unfinished TODO tasks in the description.

jnv:
I don't see this much as a blockers, Discord should be definitely in a separate PR, but if you have more tests in mind, it's better to do them now. :wink:

martinalbert:
@janhalama @jnv it's ready, I removed discord from this PR and added missing things.

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+model Message {
+	"""
+	Id
+	Identification of message
+	"""
+	id! string!
+
+	"""
+	Author
+	Author of retrieved message
```
jnv, line 75: What should I expect here? An ID? A name? An avatar URL...?

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	"""
+	author! string!
+
+	"""
+	Destination
+	Destination of retrieved message
```
jnv, line 81: Same as with input parameter. Also, do you need to return this if the user knows this upon input? :thinking:

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	threadId string
+
+	"""
+	Text
+	Content of retrieved message
+	"""
+	text string
+
+	"""
+	CreatedAt
```
jnv, line 98:
```suggestion
	Created At
```

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	Date retrieved message was created
+	"""
+	createdAt string
+
+	"""
+	CreatedAt
```
jnv, line 104:
```suggestion
	Updated At
```

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+
+	"""
+	Reactions
+	List of reactions on retrieved message
+	"""
+	reactions [
+		{
+			emoji string
+			count number
+            users [string]
```
jnv, line 117: Formatting? Also, it's not clear if this gives me a list of users' IDs or their names or whatever.

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	"""
+	updatedAt string
+
+	"""
+	Reactions
+	List of reactions on retrieved message
+	"""
+	reactions [
+		{
+			emoji string
```
jnv, line 115: I am uncertain about naming the property as "emoji", but since all platforms seem to name custom ones as "custom emojis", let's keep it. However, it should be specified what I get here; is it a text alias of the emoji like `:+1:` or `+1` or a unicode identifier?

Preferably mention this info the explanation of the `reactions` property (since I think we don't render nested properties well on the web).

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+			count number
+            users [string]
+		}
+	]
+}
+
+model Messages {
+	"""
+	PreviousPage
+	Identification of previous page
```
jnv, line 125:
```suggestion
	Cursor to the previous page for paginating backwards
```

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	"""
+	previousPage string
+	
+	"""
+	NextPage
+	Identification of next page
```
jnv, line 131: See my comments for previousPage

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	nextPage string
+
+	"""
+	Messages
+	List of retrieved messages
+	"""
+	messages [Message]
+}
+
+model Error {
```
jnv, line 142: Mind the formatting (we really should get Comlink into Prettier or something)

grid/chat/get-messages/maps/get-messages.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable jest/no-export */
+
+import { RecordingProcessOptions, SuperfaceTest } from '@superfaceai/testing';
+
+export const getMessagesTest = (
```
jnv, line 5: I'd like to see pagination in action in this test (at least for n pages)

grid/chat/get-messages/maps/mock.suma
```diff
@@ -0,0 +1,21 @@
+profile = "chat/get-messages@1.0"
+provider = "mock"
+
+"""
+GetMessages map
+"""
+map GetMessages {
```
jnv, line 7: Wow, this mock looks sad, but okay :joy:

jnv (changes requested):

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+		AfterDate
+		Date that marks beginning of messages range until current time
+		"""
+		afterDate string
+        
+		"""
+		Limit
+		Maximum number of records to retrieve
+		"""
+		limit number
```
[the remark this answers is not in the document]
martinalbert: It would be useful with slack where you can set up to 1k messages, where default is 100. I'll fix descriptions.

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+	input {
+		"""
+		Destination
+		Identification of destination to retrieve messages from
+		"""
+		destination! string!
+		
+		"""
+		AfterDate
+		Date that marks beginning of messages range until current time
```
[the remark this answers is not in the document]
martinalbert: Slack expects UNIX timestamp and discord uses filtering based on message id, so we should probably support both. Discord also returns messages with ISO format timestamp.

grid/chat/get-messages/profile.supr
martinalbert: An ID, same with destination.

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,153 @@
...
+
+	"""
+	Text
+	Content of retrieved message
+	"""
+	text string
+
+	"""
+	CreatedAt
+	Date retrieved message was created
```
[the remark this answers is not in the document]
martinalbert: Good point, I'll make it required and fix description.

grid/chat/get-messages/profile.supr
martinalbert: I'll add description for it 👍🏻

grid/chat/get-messages/profile.supr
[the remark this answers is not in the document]
martinalbert: Going through discord docs, id of each message is snowflake, which contains timestamp, so maybe supporting UNIX only will be better. 
https://discord.com/developers/docs/reference#snowflakes
https://discord.com/developers/docs/resources/channel#message-object

grid/chat/get-messages/profile.supr
martinalbert: Slack returns identifier of emoji between colons, so in case of 👍🏻 it returns `+1`

martinalbert:
@jnv I've added better description and mapping. Also, I had to go with UNIX timestamp as slack does not return ISO format for date of message creation. I'll add rate limit information and finish up discord in another PR.

grid/chat/get-messages/profile.supr
```diff
@@ -0,0 +1,187 @@
...
+	nextPage string
+
+	"""
+	Messages
+	List of retrieved messages
+	"""
+	messages [Message]
+}
+
+model Error {
```
janhalama, line 143: I miss rate limit in `Error` model.

grid/chat/get-messages/maps/slack.suma
```diff
@@ -0,0 +1,91 @@
...
+
+		response 200 "application/json" {
+			// See https://api.slack.com/methods/conversations.history#errors for more information about errors
+			return map error if (body.ok === false) {
+				title = body.error
+			}
+
+			messages = call MapMessages(messages = body.messages)
+
+			return map result if (body.has_more) {
```
janhalama, line 30: Looks like Slack dosn't support fetching previous page.

martinalbert: Yeah, I didn't find a way to link previous pages with slack. 🤷🏻‍♂️

jnv (changes requested):
I think some of my comments were not addressed yet? :thinking:

grid/chat/get-messages/maps/slack.suma
```diff
@@ -0,0 +1,106 @@
...
+				title = body.error
+			}
+		}
+	}
+}
+
+operation MapMessages {
+	messages = args.messages.filter(message => message.type === 'message')
+					
+	set if (messages.length > 0) {
```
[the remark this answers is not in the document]
martinalbert: What do you mean?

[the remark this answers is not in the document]
jnv: What I mean is that you could remove that condition completely. Or just chain it into a single call:

```js
args.messages.filter(message => message.type === 'message').map(() => ...)
```

Though more efficient approach to avoid 2 separate iterations would be to do filtering in single pass, e.g.

```js
messages = []
args.messages.forEach((message) => {
  if (message.type !== 'message') {
    return
  }
  // ...
  messages.push(result)
}
```

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+		Destination
+		Representation of a messaging destination. Can be ID or channel name, depending on the provider capability.
+		"""
+		destination! string!
+		
+		"""
+		After Date
+		Retrieve only messages after this date and time, UNIX timestamp format (e.g. 1095379198.75) 
+		"""
+		afterDate number
```
jnv, line 20: I think we discussed that calling it Date is misleading since it expects a timestamp, no?

```suggestion
		"""
		After Timestamp
		Retrieve only messages after the given UNIX timestamp (in seconds, e.g. 1095379198.75) 
		"""
		afterTimestamp number
```

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+		
+		"""
+		Before Date
+		Retrieve only messages before this date and time, UNIX timestamp format (e.g. 1095379198.75)
+		"""
+		beforeDate number
```
jnv, line 26: Same as with afterDate

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+}
+
+model Attachment {
+	"""
+	Id
+	Unique identifier of the attachment
+	"""
+	id! string!
```
jnv, line 90: I wouldn't require an attachment's ID as mandatory field.

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+
+	"""
+	Created At
+	UNIX timestamp (e.g. 1095379198.75) when the attachment was created
+	"""
+	createdAt! number!
```
jnv, line 96: I am somewhat torn here, we use `createdAt` for ISO date fields in other profiles, so maybe use `createdTimestamp` instead?

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+
+	"""
+	File Name
+	Name of attached file
+	"""
+	fileName string
+
+	"""
+	Media Type
+	Format of the attached file
```
jnv, line 106: Would be useful to mention what to expect here; MIME type? String like "video", "image"? Or it's provider specific?

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+
+	"""
+	Url
+	Source url of the attachment
+	"""
+	url string
+
+	"""
+	Preview
+	Preview of the attached file
```
jnv, line 118: What is this? An URL with the preview? Is it preview image? If this is the case, I suggest renaming this to `previewUrl`

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+	author! {
+		id! string!
+		username string!
+	}!
+
+	"""
+	Created At
+	UNIX timestamp (e.g. 1095379198.75) when the message was created
+	"""
+	createdAt! number!
```
jnv, line 143: Again with createdAt :thinking:

grid/chat/messages/profile.supr
```diff
@@ -0,0 +1,278 @@
...
+	Text
+	Contents of the retrieved message
+	"""
+	text string
+
+	"""
+	Updated At
+	UNIX timestamp (e.g. 1095379198.75) when the message was last updated
+	"""
+	updatedAt number
```
jnv, line 173: updatedTimestamp?

grid/chat/messages/maps/slack.suma
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	return details
+}
+
+operation ComposeURL {
+	timestamp = args.ts
+	destination = args.dest
+
+	return `https://slack.com/api/conversations.history?channel=${destination}&latest=${timestamp}&inclusive=true&limit=1`
```
jnv, line 175: Huh, if this is publicly exposed URL, I don't think it should point to an API, but maybe I misunderstand how this is used?

jnv (changes requested):
A few notes regarding docstring.

I am a bit torn on `*At` and `*Timestamp` but I am probably okay with using `createdAt` and `updatedAt` with timestamps.

grid/chat/messages/profile.supr
martinalbert: This is implemented in PR with discord, I'll add it here as well.

grid/chat/messages/profile.supr
martinalbert: Both providers return id of attachments 🤷🏻‍♂️

grid/chat/messages/profile.supr
martinalbert: I think suffix `At` is general enough to be referencing to timestamp as well.

grid/chat/messages/profile.supr
martinalbert: This only applies to slack and it contains string preview of attachment file.

grid/chat/messages/maps/slack.suma
martinalbert: Yeah, this points to the API 😕

grid/chat/messages/profile.supr
martinalbert: It's also used in other capabilities in station.

grid/chat/messages/profile.supr
jnv: Hmm, okay, so keep it, perhaps we will need it for some consecutive use-cases.

grid/chat/messages/profile.supr
jnv: Okay, so looking at docs: https://api.slack.com/types/file#previews
Seems this has limited usage for text snippets. I wouldn't mind removing it. Alternatively please expand the description like "Short text preview of text attachments (e.g. snippets, posts)"

martinalbert: I agree, I'll remove it.

grid/chat/get-messages/maps/slack.suma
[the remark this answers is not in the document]
martinalbert: That looks good, I'll refactor it.

grid/chat/messages/profile.supr
martinalbert: It's MIME type in both providers, will this be enough? 👇🏻 

```suggestion
	MIME type of the attached file.
```

grid/chat/get-messages/maps/slack.suma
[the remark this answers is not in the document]
martinalbert: Unfortunately jessie expression such as `args.messages.forEach` can't be used outside assignment, so I had to chain it, but when I chaned it, language server had problem with that, so I just removed the condition.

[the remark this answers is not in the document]
martinalbert: Ok it no longer throws that error, so I chained it.

martinalbert:
@janhalama @jnv I added some review fixes and also refactored mapping due to performance issues.

grid/chat/messages/profile.supr
jnv: Great, thanks!

package.json
```diff
@@ -22,7 +22,7 @@
   },
   "devDependencies": {
     "@superfaceai/ast": "^1.0.0",
-    "@superfaceai/cli": "^1.0.1",
+    "@superfaceai/cli": "1.1.0-beta.0",
```
jnv, line 25: Is this change necessary for this PR, can't it be done in a separate PR?

grid/chat/messages/maps/slack.suma
```diff
@@ -0,0 +1,230 @@
...
+        fileName: file.name,
+        mediaType: file.mimetype,
+        url: file.permalink_public,
+        preview: file.preview,
+      }));
+
+      resultMessage.attachments = attachments;
+    }
+
+    resultMessage.url = `https://slack.com/api/conversations.history?channel=${args.destination}&latest=${message.ts}&inclusive=true&limit=1`;
```
jnv, line 116: Okay, so this is last outstanding issue for me.

I see that Slack has a dedicated method for that, which would be costly to run for every message: https://api.slack.com/methods/chat.getPermalink

But I'd say consumer's expectation here is to return a URL which is presentable to the end user and take them to the relevant message. So returning this is in my opinion worse than returning nothing at all.

What is missing to be able to construct this URL manually. For example here is threaded message:

```
https://superfaceai.slack.com/archives/C013602CL2Y/p1645175948429299?thread_ts=1645043940.215569&cid=C013602CL2Y
```

I don't think we have any workspace identifier (or do we?), but we do have channel ID, message ID and thread ID.

jnv (changes requested):
Message url is a blocker for me.

package.json
martinalbert: It's mainly to unblock lint in CI 🤷🏻‍♂️

grid/chat/messages/maps/slack.suma
martinalbert: I'll try to create public URL then. 🙏🏻

grid/chat/messages/maps/slack.suma
```diff
@@ -0,0 +1,231 @@
+profile = "chat/messages@1.0"
+provider = "slack"
+
+"""
+GetMessages map
+"""
+map GetMessages {
+	http GET "/conversations.history" {
```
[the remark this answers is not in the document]
martinalbert: I updated formatting in all PRs to 2 space indentation.

grid/chat/messages/maps/slack.suma
martinalbert: I would also need workspace name to compose correct URL, so it's not possible without that call.

We could return partially generated url, user would have to add his workspace name. For example, I'm able to generate URL: `https://slack.com/archives/CF3H7S63W/p1644310634546009`, but to make it work, I have to add workspace name before domain i.e. `https://<workspace-name>.slack.com/archives/CF3H7S63W/p1644310634546009`. URL for threads are more complicated, but still need workspace name.

jnv:
Ok to go, let's rather add URLs in a separate PR.

PR #188, opened 2022-01-28
feat(crm): Track Contact Event

kysely:
## Description

Adding `crm/track-contact-event` profile with a map to Crisp.

## Motivation and Context

This is different from the existing [`analytics/track-event`](https://superface.ai/analytics/track-event) as it's scoped for the specific contact whereas the existing use case is oriented to more general analytics systems.

Rendered as

preview

## Types of changes

## Checklist:


grid/crm/track-contact-event/maps/crisp.suma
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    response 201 "application/json" {
+      return map result if (body.reason === "added") {
+        contactId = input.contactId
+      }
+
+      map error {
+        title = "Unexpected error"
+        detail = body
```
kysely, line 26: Unsure how this assignment will be passed to the user

grid/crm/track-contact-event/profile.supr
```diff
@@ -0,0 +1,78 @@
...
+name = "crm/track-contact-event"
+version = "1.0.0"
+
+"""
+Track Contact Event
+Sends an event data to CRM for a specific contact
+"""
+usecase TrackContactEvent unsafe {
+  input {
+    contactId!
```
jnv, line 14: Would be helpful to document this property, ie. should I use an email or some provider-specific ID?

grid/crm/track-contact-event/maps/crisp.suma
jnv: You have specified the property type as string, so I am worried that passing in an object will cause a validaton error.

janhalama (commented):
Looks good to me. The only think that should be fixed is the mapping of unexpected error.

grid/crm/track-contact-event/maps/crisp.suma
kysely: @jnv @janhalama Since this is inside `application/json` handler already, I bet on the response actually being a JSON object and am directly stringifying it. Thanks!

janhalama (approved):
👍

PR #189, opened 2022-01-31
Add rate limit into `social-media/followers` profile

janhalama:
## Description

This PR adds rate limit information into GetFollowers use-case result and error outcome. It also adds machine processable error codes to the error model.

`social-media/followers` version bumped to 1.1.0

## Types of changes

## Checklist:


grid/social-media/followers/profile.supr
```diff
@@ -51,6 +51,32 @@ usecase GetFollowers safe {
       """
       imageUrl string
     } ]
+    """
+    Rate limit
+    Rate limit details.
+    """
+    rateLimit {
+      """
+      Total requests available
```
janhalama, line 60: The confusing think in some profile use-cases might be, that one-sdk perform will actually send more than 1 request to the Twitter API endpoints.

grid/social-media/followers/profile.supr
```diff
@@ -64,6 +90,29 @@ usecase GetFollowers safe {
     A human-readable explanation specific to this occurrence of the problem.
     """
     detail
+
+    """
+    Code
+    Error code.
+    """
+    code! ErrorCode!
```
janhalama, line 98: I have added machine processable error code. Code is marked as required and non-nullable.

grid/social-media/followers/profile.supr
```diff
@@ -51,6 +51,32 @@ usecase GetFollowers safe {
       """
       imageUrl string
     } ]
+    """
+    Rate limit
+    Rate limit details.
+    """
+    rateLimit {
```
janhalama, line 58: Harmonization of rate limiting across providers will be hard. Some of them work with rolling window, some do not provide exact limit number.

grid/social-media/followers/profile.supr
jnv: Yeah, so we could either try to be smart about this and somehow try to guess the number of remaining requests for this use case, or just mention this caveat in the documentation: 1 use-case perform doesn't equal 1 request.

grid/social-media/followers/profile.supr
```diff
@@ -51,6 +51,32 @@ usecase GetFollowers safe {
...
+      """
+      Remaining requests available
+      Remaining requests available in the time window.
+      """
+      remainingRequestsAvailable number
+      """
+      Remaining requests available in percents
+      Remaining requests available in the time window in percents.
+      """
+      remainingRequestsAvailableInPercent number
```
jnv, line 73: It's not shorter but maybe more readable?

```suggestion
      remainingRequestsAvailablePercentage number
```

grid/social-media/followers/profile.supr
```diff
@@ -93,3 +142,11 @@ usecase GetFollowers safe {
     }
   }
 }
+
+model ErrorCode enum {
```
jnv, line 146: BTW if we expand this enum, will it require a minor version bump? Technically I think it does.

grid/social-media/followers/profile.supr
```diff
@@ -64,6 +90,29 @@ usecase GetFollowers safe {
...
+    Code
+    Error code.
+    """
+    code! ErrorCode!
+
+    """
+    Rate limit
+    Rate limit error details.
+    """
+    rateLimit {
```
jnv, line 104: Shouldn't this share the same model with `rateLimit` in result?

grid/social-media/followers/profile.supr
jnv: I see the biggest challenge with use-cases performing multiple requests in a single perform _and_ different limits per endpoint.

jnv (commented):
Good start!

I suggest adding one more property to rateLimit properties: `bucket` or `realm` or similar, and `scope`. This is inspired by [Discord's rate limit headers](https://discord.com/developers/docs/topics/rate-limits)

- `bucket` identifies parts of the API with dedicated rate limits, see https://developer.twitter.com/en/docs/twitter-api/rate-limits#headers-and-codes (e.g. `follows` in this usecase)
- `scope` whether this is global, for app, or for user; however I don't know if we can identify this since the bearer token can be both for authorized user, or for app, so I am okay omitting it for now.

grid/social-media/followers/profile.supr
janhalama: We return `ErrorCode` enum as result of perform call. Expanding the enum could break client code see:

```
switch (error.code) {
    case 'Unauthenticated':
        // handle unauthenticated error
        break;
    case 'Unauthorized':
        // handle unauthorized error
        break;
    default: // unknown error code values
        throw new Error('Not implemented');
}
```

I would bump major version of ErrorCode enum.

---

If the enum is used only in inputs minor version bump is sufficient.

grid/social-media/followers/profile.supr
janhalama: > we could either try to be smart about this and somehow try to guess the number of remaining requests for this use case
I would go with this option. In case of Twitter we have enough information to calculate it.

> or just mention this caveat in the documentation: 1 use-case perform doesn't equal 1 request.
In this case we could also add number of API requests the map calls.

grid/social-media/followers/profile.supr
janhalama: > I see the biggest challenge with use-cases performing multiple requests in a single perform and different limits per endpoint.

I see two options:

1. Instead of returning remaining requests send to the client remaining performs of the use-case and calculate the worst case.

But I do not know what to do in case of subsequent fetches of referenced entities (e. g. attachments) with variable count.

2. Return remaining requests in array of buckets (brings complexity to the client)

```
rateLimits: [
 {
  bucket: 'followers',
  requestsPerUsecase: 1,
  remainingRequestsAvailable: 100,
  ...
 },
 {
  bucket: 'attachments',
  requestsPerUsecase: undefined, //we do not know the number
  remainingRequestsAvailable: 10,
  ...
 }
]
```

grid/social-media/followers/profile.supr
```diff
@@ -51,6 +51,32 @@ usecase GetFollowers safe {
...
+      """
+      Total requests available
+      Total requests available in the time window.
+      """
+      totalRequestsAvailable number
+      """
+      Remaining requests available
+      Remaining requests available in the time window.
+      """
+      remainingRequestsAvailable number
```
jnv, line 68:
```suggestion
      remainingRequests number
```

grid/social-media/followers/profile.supr
```diff
@@ -51,6 +51,32 @@ usecase GetFollowers safe {
...
+    """
+    Rate limit
+    Rate limit details.
+    """
+    rateLimit {
+      """
+      Total requests available
+      Total requests available in the time window.
+      """
+      totalRequestsAvailable number
```
jnv, line 63:
```suggestion
      totalRequests number
```

grid/social-media/followers/profile.supr
```diff
@@ -51,6 +51,32 @@ usecase GetFollowers safe {
...
+      """
+      Remaining requests available in percents
+      Remaining requests available in the time window in percents.
+      """
+      remainingRequestsAvailableInPercent number
+      """
+      Reset date time
+      Date and time when the rate limits will reset (in ISO 8601 format). Available only for providers without rolling rate limit window.
+      """
+      resetDateTime string
```
jnv, line 78: You will receive this as Unix timestamp which, AFAIK, we can't easily format to ISO date in map (heads up @TheEdward162). I'd just pass the timestamp (converted from string to number):
```suggestion
      resetTimestamp number
```

janhalama: changed

grid/social-media/followers/profile.supr
janhalama: changed

grid/social-media/followers/profile.supr
janhalama: changed

grid/social-media/followers/profile.supr
janhalama: Good point. `RateLimit` model is shared in error and result outcomes.

grid/social-media/followers/profile.supr
janhalama: changed

grid/social-media/followers/profile.supr
janhalama: We will not get into this edge case with Twitter. We will solve it once necessary by updating major version or splitting the use-case.

grid/social-media/followers/profile.supr
janhalama: I added relevant Twitter error codes to the profile.

grid/social-media/followers/maps/twitter.suma
```diff
@@ -77,3 +56,89 @@ operation MapFollower {
...
+  set if(totalRequests > 0 && remainingRequests !== undefined) {
+    remainingRequestsPercentage = remainingRequests / totalRequests * 100
+  }
+
+  set if(resetTimestampHeader && parseInt(resetTimestampHeader, 10) !== 'NaN') {
+    resetTimestamp = parseInt(resetTimestampHeader, 10)
+  }  
+
+  return {
+    bucket = 'get-users-followed-lists'
```
jnv, line 88: Hmm, I don't quite follow the relation of this bucket name and the docs nor endpoint; how about just copying it verbatim from the docs as `Follows lookup`?

```suggestion
    bucket = 'Follows lookup'
```

(or if you really don't like spaces)

```suggestion
    bucket = 'follows-lookup'
```

grid/social-media/followers/profile.supr
```diff
@@ -89,7 +113,78 @@ usecase GetFollowers safe {
...
+  Unknown error
+  Returned when an unexpected error occured.
+  """
+  UnknownError
+}
+
+model RateLimit {
+  """
+  Bucket
+  Bucket identifies parts of the API with dedicated rate limits.
```
jnv, line 167:
```suggestion
  Different parts of API may have different rate limits.
  Bucket identifies to which part of API the rate limits apply.
```

jnv (approved):
A few minor notes, LGTM otherwise.

grid/social-media/followers/maps/twitter.suma
janhalama: That was my mistake. I misinterpreted Twitter rate limit documentation. 👍

PR #191, opened 2022-02-02
Add rate limit to `social-media/posts-lookup` profile

janhalama:
## Description

This PR
- adds rate limit support to `social-media/posts-lookup` profile
- bumps minor version of the profile to 1.2.0
- adds `social-media/posts-lookup` profile examples
- adds support for  `social-media/posts-lookup` v1.2.0 to Twitter and Mock map

## Types of changes

## Checklist:


grid/social-media/posts-lookup/maps/twitter.suma
```diff
@@ -196,3 +260,11 @@ operation MapUser {
 
   return author
 }
+
+operation GetSearchRecentTweetsBucket {
+  return "Search Recent Tweets"
```
jnv, line 265: Watch out for bucket's string capitalization, for consistency. Assuming these buckets are used programmatically, they must be consistent across maps from single provider.
```suggestion
  return "Search recent tweets"
```

grid/social-media/posts-lookup/maps/twitter.suma
```diff
@@ -155,32 +161,90 @@ operation MapMediaAttachment {
...
+    remainingRequests = parseInt(remainingRequestsHeader, 10)
+  }
+
+  set if(totalRequests > 0 && remainingRequests !== undefined) {
+    remainingRequestsPercentage = remainingRequests / totalRequests * 100
+  }
+
+  set if(resetTimestampHeader && parseInt(resetTimestampHeader, 10) !== 'NaN') {
+    resetTimestamp = parseInt(resetTimestampHeader, 10)
+  }
```
jnv, line 189: I am thinking about this duplicated `parseInt`. How about parsing the headers directly and then only check for NaN? I _think_ `parseInt` never throws, it just gives you NaN or a garbage number.

```
totalRequestsHeader = args.headers['x-rate-limit-limit']
// ...
set if(totalRequestsHeader !== 'NaN') {
  totalRequests = totalRequestsHeader
}
```

Also not sure if NaN is handled correctly; can we use [Number.isNaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) or straight `isNaN` instead?

jnv (commented):
In general okay, but I have concern about NaN comparison I haven't realized in previous PR.

grid/social-media/posts-lookup/maps/twitter.suma
janhalama: OK I will go with lover case. First I wanted to use exactly the same value as is the name of resource documented in Twitter rate limit table. But in case of search there are two levels:
image

grid/social-media/posts-lookup/maps/twitter.suma
janhalama: Good points 👍

janhalama: To be honest I did not test the NaN edge case. I would like to have ability to write unit tests for Comlink -> mock inputs.

grid/social-media/posts-lookup/maps/twitter.suma
janhalama: fixed

grid/social-media/posts-lookup/maps/twitter.suma
janhalama: fixed

janhalama:
@jnv thanks for review, I fixed your comments

jnv (approved):
LGTM, mind that the invalid NaN comparisons are already in the followers map.

PR #193, opened 2022-02-03
feat: Add discord map for `chat/messages` profile

martinalbert:
## Description
Adds discord provider along with the map.

## Types of changes

## Checklist:


grid/chat/get-messages/maps/discord.suma
```diff
@@ -0,0 +1,203 @@
...
+			id: message.id,
+			author: {
+				id: message.author.id,
+				username: message.author.username
+			},
+			createdAt: Date.parse(message.timestamp),
+			text: message.content
+		}
+		const discordEpoch = 1420070400000;
+		const getTimestamp = (snowflake) => {
```
janhalama, line 112: There is an operation GetTimestamp defined, is not this duplicity?

grid/chat/get-messages/profile.supr
```diff
@@ -14,10 +14,18 @@ usecase GetMessages {
 		destination! string!
 		
 		"""
-		AfterDate
-		Retrieve only messages after this date and time, UNIX timestamp format (e.g. 1095379198.75)
+		After Date
+		Retrieve only messages after this date and time, UNIX timestamp format (e.g. 1095379198.75) 
+		or snowflake format, which includes UNIX timestamp in its first 42 bits.
```
janhalama, line 19: Why do we need snowflake format here? It is provider specific. Do we return snowflake in the output to the client?

grid/chat/get-messages/maps/discord.suma
```diff
@@ -0,0 +1,203 @@
...
+More about snowflakes: https://discord.com/developers/docs/reference#snowflakes
+"""
+operation GetSnowflake {
+	discordEpoch = 1420070400000
+
+	date = args.date
+	processedDate = date - discordEpoch
+
+	// if date is not a string, it should not have property `length`
+	return if (date.length === undefined) (BigInt(processedDate) << BigInt(22)).toString()
```
janhalama, line 183: Using the length property doesn't look very convenient. Unix timestamp should be passed as number and shownflake as string? `isNaN` can not be used?

grid/chat/get-messages/maps/discord.suma
```diff
@@ -0,0 +1,203 @@
...
+
+			map result {
+				previousPage = messages[0].id
+				nextPage = messages[messages.length - 1].id
+				messages = messages
+				rateLimit = rateLimit
+			}
+		}
+
+		response 400 "application/json" {
```
janhalama, line 45: I miss rate limit info in error outcomes.

janhalama (commented):
Overal looks good. Please 🙏 fix the indentation, there are format issues all over the code. I am not sure if the option to pass snowflake as beforeDate and afterDate is needed.

grid/chat/get-messages/maps/discord.suma
martinalbert: I'll refactor it

grid/chat/get-messages/profile.supr
martinalbert: `previousPage` and `nextPage` are snowflakes when used with discord, so yes it is provider specific. I did this to enable pagination forwards and backwards at least for discord.

grid/chat/get-messages/maps/discord.suma
martinalbert: True, I don't like this either. We can remove it if we limit discord to forwards only pagination.

martinalbert:
Thanks @janhalama, I left some comments in review.

grid/chat/get-messages/profile.supr
martinalbert: Since we can't properly check input types, I'll limit discord to forward pagination only and remove `string` type. @jnv are you okay with it?

grid/chat/get-messages/maps/discord.suma
janhalama: I would add it to every error response.

grid/chat/messages/maps/discord.suma
```diff
@@ -0,0 +1,276 @@
...
+
+"""
+Operation MapMessages
+Returns array mapping messages in received body
+"""
+operation MapMessages {
+	listOfMessages = []
+	
+	// workaround for map interpreter variables merging issue
+	listOfMessages = call foreach(index of Array(args.messages.length).fill().map((_, i) => i)) MapMessage(
```
janhalama, line 122: @martinalbert This will cause the same performance  problems we had in `social-media/posts-lookup` Twitter map. See my PR https://github.com/superfaceai/station/pull/207/files. The solution is to use single Jessie script for response mapping.

janhalama: Not exactly the workaround, the problem is that OneSDK would have to do thousands of Jessie script evals and the overhead with initialisation of Jessie script eval is high.

martinalbert: Thanks! I'll refactor it.

martinalbert: I wonder how would sdk perform if there would be a variable instead of raw `Array(args)...` expression. 🤔

janhalama: > I wonder how would sdk perform if there would be a variable instead of raw Array(args)... expression. 🤔

IMO there won't be any difference in the performance.

martinalbert: @janhalama will you solution work when I iterate on top of array of returned items in body or it's important to use `[body].map(...`?

janhalama: You do not need to access properties from outside of the array you do not need to use the `[body].map(...` construct.

This should work well:
```
listOfMessages = args.messages.map(message, => {
...
}
```

martinalbert: Alright, thanks

martinalbert: @janhalama I refactored it, check it out if it's good enough please.

janhalama: looks good

janhalama (approved):
Looks good to me 👍

grid/chat/messages/maps/discord.suma
```diff
@@ -0,0 +1,253 @@
...
+      });
+
+      resultMessage.attachments = attachments;
+    }
+
+    if (message.message_reference && message.message_reference.message_id) {
+      resultMessage.parentId = message.message_reference.message_id;
+    }
+
+    resultMessage.url = `https://discord.com/api/channels/${message.channel_id}/messages/${message.id}`;
```
jnv, line 121: Again, this should be a publicly accessible URL.

grid/chat/messages/maps/discord.suma
```diff
@@ -0,0 +1,253 @@
...
+    }
+  }
+}
+
+"""
+Operation MapMessages
+Returns array mapping messages in received body
+"""
+operation MapMessages {
+  listOfMessages = []
```
jnv, line 61: This is redundant.

jnv (changes requested):

grid/chat/messages/maps/discord.suma
martinalbert: Removed it

martinalbert: Removed it

jnv (approved):

grid/chat/messages/maps/discord.suma
jnv: @martinalbert Please fix this one thing.

PR #194, opened 2022-02-03
feat: add `chat/channels` with discord and slack map

martinalbert:
## Description
Adds new profile `chat/channels` with two new maps.

## Motivation and Context
Need of channel id for profile `chat/messages`.

## Types of changes

## Checklist:


grid/chat/channels/maps/channels.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable jest/no-export */
+
+import { RecordingProcessOptions, SuperfaceTest } from '@superfaceai/testing';
+
+export const getChannelsTest = (
+  provider: string,
+  options?: { server?: string; recordingOptions?: RecordingProcessOptions }
```
janhalama, line 7: I did not find use-case to retrieve servers. Are we going to add another profile for it?

grid/chat/channels/maps/discord.suma
```diff
@@ -0,0 +1,183 @@
+profile = "chat/channels@1.0"
+provider = "discord"
+
+"""
+GetChannels map
+"""
+map GetChannels {
+    return map error if (input.server === undefined) {
```
janhalama: nitpicking: part of the Comlink map uses 4 spaces / 2 spaces for indentation

janhalama (commented):
Looks good 👍 

 🙏 mind indentation it is hard to review the code

grid/chat/channels/maps/channels.ts
martinalbert: I had to implement it for messaging demo, so I can open PR for it 🙌🏻

grid/chat/channels/maps/discord.suma
martinalbert: I already updated indentation. Hopefully it's better now.

janhalama (approved):
👍

grid/chat/channels/maps/discord.suma
```diff
@@ -0,0 +1,183 @@
...
+}
+
+"""
+Operation MapChannels
+Returns array mapping channels in received body
+"""
+operation MapChannels {
+  channels = []
+
+  channels = args.channels.filter(channel => [0, 1, 3, 5].includes(channel.type)).map(channel => {
```
jnv, line 48: I don't see how this filter translates to public/private type filter; in fact, I don't see that input being used anywhere.

I see this filter/map pattern repeating in other of your maps as well, but I don't like it because it causes you to iterate the array twice. Have you considered using `forEach` with inlined filter?

Also I wanted to suggest using Set instead of `[].includes` but then I realized we don't allow `new` :facepalm:

grid/chat/channels/maps/discord.suma
```diff
@@ -0,0 +1,183 @@
...
+    const discordEpoch = 1420070400000;
+    const getTimestamp = (snowflake) => Number(BigInt(snowflake) >> BigInt(22)) + discordEpoch;
```
jnv, line 50: Can you construct that outside of the map function?

```
channels = []
getTimestamp = (snowflake) => Number(BigInt(snowflake) >> BigInt(22)) + 1420070400000;
```

Also it would be helpful to add a comment with link to Discord docs for this algorithm

grid/chat/channels/profile.supr
```diff
@@ -0,0 +1,191 @@
...
+
+"""
+Get Channels
+Retrieve paginated list of channels, e.g. text channels in slack or discord
+"""
+usecase GetChannels {
+  input {
+    """
+    Server
+    Representation of a server. Can be guild or server, depending on the provider capability.
```
jnv, line 12: Did we establish using `server` as the identifier here? I think in most Slack-like applications it's called "workspace" (I know about Slack, Rocket.Chat, Mattermost). Calling it "server" can be also confusing for apps supporting self-hosted deployments _and_ workspaces. So I'd consider renaming this to `workspace` or perhaps more abstract `space`. But maybe @janhalama has a different opinion here?

grid/chat/channels/profile.supr
```diff
@@ -0,0 +1,191 @@
...
+    """
+    server string
+
+    """
+    Types
+    List of categories of channels
```
jnv, line 18: Does it list all channels if I omit this?

I am also somewhat concerned about a broad meaning of "type"; perhaps I might want to filter text vs. voice channels on Discord or something else. Consider renaming this to e.g. `visibility` or `privacy`

Documentation improvement:

```suggestion
    Channel Types Filter
    Limit the listing to either public or private channels; all channels are listed by default.
```

grid/chat/channels/profile.supr
```diff
@@ -0,0 +1,191 @@
...
+  result {
+    """
+    Channels
+    List of retrieved channels
+    """
+    channels! [Channel]
+
+    """
+    Next Page
+    Cursor to the next page for pagination forwards
```
jnv, line 49:
```suggestion
    Cursor to the next page of listing
```

grid/chat/channels/profile.supr
```diff
@@ -0,0 +1,191 @@
...
+
+  """
+  Created At
+  UNIX timestamp (e.g. 1095379198.75) when the channel was created
+  """
+  createdAt! number!
+
+  """
+  Name
+  Name of retrieved channel
```
jnv, line 111:
```suggestion
  Name of the retrieved channel
```

grid/chat/channels/profile.supr
```diff
@@ -0,0 +1,191 @@
...
+  Description
+  Description of the channel. Can describe purpose or use of the channel.
+  """
+  description string
+
+  """
+  Num Of Member
+  Number of members in retrieved channel
+  """
+  numOfMembers number
```
jnv, line 125: I'd avoid shortening in fields' names as much as possible. I suggest renaming this to `membersCount`

```suggestion
  """
  Members Count
  Number of members of the retrieved channel
  """
  memebersCount number
```

grid/chat/channels/maps/slack.suma
```diff
@@ -0,0 +1,212 @@
...
+        statusCode = statusCode,
+        error = body,
+        headers = headers,
+      ) if (!body.ok)
+
+      return map error if (!body.ok) error
+
+      channels = call MapChannels(channels = body.channels)
+      rateLimit = call MapRateLimitDetails(headers = headers)
+      result.channels = channels
```
jnv, line 35: Why not assign result.channels directly? Also am I mad or I don't see any `result` constructor? Is that some implicit Comlink feature? :grimacing:

grid/chat/channels/maps/slack.suma
```diff
@@ -0,0 +1,212 @@
...
+    if (channel.topic.value && channel.topic.value !== '') {
+      resultChannel.description = channel.topic.value;
+    }
+
+    if (channel.purpose.value && channel.purpose.value !== '') {
+      resultChannel.description = 
+        resultChannel.description === undefined 
+        ? channel.purpose.value
+        : `${resultChannel.description} - ${channel.purpose.value}`;
+    }
```
jnv, line 82: I am not 100% sure about this creative concatenation of description and purpose, personally I'd use purpose only as a fallback when description is not present.

grid/chat/channels/maps/slack.suma
```diff
@@ -0,0 +1,212 @@
...
+
+    return resultChannel;
+  });
+
+  return channels
+}
+
+operation typesToQuery {
+  types = args.types.map(type => {
```
jnv, line 95: So what happens if I give you `undefined` instead of array for `types`?

grid/chat/channels/maps/discord.suma
```diff
@@ -0,0 +1,183 @@
+profile = "chat/channels@1.0"
+provider = "discord"
+
+"""
+GetChannels map
+"""
+map GetChannels {
+  return map error if (input.server === undefined) {
+    title = "Input parameter `server` is required in case of discord provider"
```
jnv, line 9:
```suggestion
    title = "Input parameter `server` is required for provider discord"
```

jnv (changes requested):
Multiple concerns, please start with comments on profile.

grid/chat/channels/maps/discord.suma
martinalbert: Private/public is relevant only to slack. 😕 

I like `forEach` approach, but I can't use it outside assignment, so that's why I wrote it this way.

grid/chat/channels/maps/discord.suma
martinalbert: Not possible 😢  I'll comment it more

grid/chat/channels/maps/slack.suma
martinalbert: Yeah it's a feature 😄 This is needed because of resolving objects is not really finished in linter, when one object switches from jessie expression to comlink object and vice versa, it doesn't keep properties from first type of assignment.

grid/chat/channels/maps/slack.suma
martinalbert: Check out snapshots, I think it's useful, but I don't have strong opinion on this as well.

grid/chat/channels/profile.supr
janhalama: I agree that server is not the best option. I would go with `workspace`.

grid/chat/channels/maps/discord.suma
jnv: Or actually use reduce, it makes the most sense. Just do mutate an accumulator to avoid generating too much garbage.

grid/chat/channels/profile.supr
martinalbert: Alright, I will also open new PR with `chat/workspaces` profile for getting workspaces (and guilds in case of discord).

grid/chat/channels/profile.supr
martinalbert: Good point. In case of slack, it does list all channels if omitted, but it list just relevant channels in case of discord. Discord does not have any filtering parameters when listing guild channels, so we can only filter channel types after we get them.

martinalbert: We can either extend this to contain relevant filtering types such as `text` or `voice` for discord and handle them in map, or create another filtering property for discord types. 🤷🏻‍♂️

jnv (approved):

PR #196, opened 2022-02-07
feat: oauth/refresh-token

jnv:
## Description

This implements OAuth 2.0 token refresh for Google APIs.

While we plan to have support for this in SDK, we already need to use this in PoCs now. Plus it's easier to use and better decoupled than handling this through Passport.js strategy.

The worst thing here is redacting the test recordings because:

- use case accepts sensitive inputs
- which are url encoded and therefore not matched by the testing lib
- and I need to remove access token from the response

Thanks to @martinalbert for help!

## Types of changes

## Checklist:


grid/oauth/refresh-token/profile.supr
```diff
@@ -0,0 +1,70 @@
...
+    clientId string!
+
+    """
+    Client Secret
+    Client authentication; may be also accepted as provider parameter.
+    """
+    clientSecret string!
+  }
+
+  result {
```
janhalama, line 30: According to spec auth server may issue new refresh token, we should add it as optional to the result.

https://datatracker.ietf.org/doc/html/rfc6749#section-6

janhalama (approved):
LGTM 👍 

I would add refresh token to the result as mentioned in comment. Otherwise looks good.

grid/oauth/refresh-token/profile.supr
jnv: Yes, I have noticed that Twitter precisely does that, I am covering this in a follow-up PR.

PR #197, opened 2022-02-08
feat(oauth/refresh-token): Add twitter provider

jnv:
## Description

Follow up to #196. This adds the Twitter as a oauth/refresh-token provider in addition to these changes:

- Profile also returns `refreshToken` which Twitter issues on every refresh
- Tests redact the `refresh_token` in the response
- Since new refresh token invalidates a previous one, I have settled with shared mutable object (`currentInput`) in tests to pass the refresh token between tests; it's not pretty so if you have a better solution, I am up for it
- Also the Twitter test logs out `TWITTER_REFRESH_TOKEN` on recording so an updated value can be pasted into `.env`

Technically adding a new property to result is a minor change, but I'd like to merge this along with #196 which isn't yet released, so there's no need for version change.

## Types of changes

## Checklist:


janhalama:
> I'd like to merge this along with https://github.com/superfaceai/station/pull/196 which isn't yet released, so there's no need for version change.

Publishing is triggered by GA after merging to main. How are you goign to avoid running publish action for https://github.com/superfaceai/station/pull/196?

jnv:
Alternatively I could just cancel the CI action once I merge #196...

janhalama:
> Will the publishing of 196 prevent me from publishing this one?

Yes, Brain response will be:

image

Merging this into 196 or or canceling GA action will work.. I am OK with it.

grid/oauth/refresh-token/maps/twitter.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const input = {
+  refreshToken: process.env.TWITTER_REFRESH_TOKEN,
+  clientId: process.env.TWITTER_CLIENT_ID,
+  clientSecret: process.env.TWITTER_CLIENT_SECRET,
+};
+
+refreshTokenTest('twitter', input, result => {
+  if (result.refreshToken !== REFRESH_TOKEN_REDACTED_VALUE) {
+    console.log('Latest Twitter refresh token:');
```
janhalama, line 15: Why do we need to log refresh token here?

providers/twitter.json
```diff
@@ -10,6 +10,14 @@
     {
       "name": "accessToken",
       "description": "User-authorized OAuth 2.0 access token"
+    },
+    {
```
janhalama, line 14: It will become unclear for the user which use-case requires which integration parameters. But I do not have solution in mind for it.

janhalama (approved):
LGTM

grid/oauth/refresh-token/maps/twitter.test.ts
jnv: In case you want to rerun the recording, you need to paste the latest refresh token manually. This is a crude solution but IMO it's better than nothing.

providers/twitter.json
jnv: Yeah, agreed. Start a beglog for that?

PR #200, opened 2022-02-08
fix(twitter): Revert twitter provider parameters

jnv:
This reverts integration parameters changes to Twitter provider introduced in #197. Providers are immutable and cannot be updated just like this. Additionally these parameters are (so far) used only by oauth2/refresh-token profile, which might be confusing in context of other profiles.

Currently integration parameters can be passed to the map even when not specified in provider JSON, so I am keeping this behavior with the existing refresh-token map. The cleaner (though not universal) solution here would be to use default inputs.

### DIFF
--- a/providers/twitter.json
+++ b/providers/twitter.json
@@ -10,14 +10,6 @@
     {
       "name": "accessToken",
       "description": "User-authorized OAuth 2.0 access token"
-    },
-    {
-      "name": "clientId",
-      "description": "Client ID of the registered OAuth application"
-    },
-    {
-      "name": "clientSecret",
-      "description": "Client Secret of the registered OAuth application"
     }
   ],
   "defaultService": "default"


janhalama (approved):
LGTM

One more reason to not add client id and secret to provider integration parameters is ongoing work on native support for OAuth 2.0 refresh token grant in OneSDK. See: https://github.com/superfaceai/one-sdk-js/pull/195

jnv:
@janhalama Good point, I can totally see a possibility to move access token to a dedicated security (?) parameter as well.

janhalama:
> @janhalama Good point, I can totally see a possibility to move access token to a dedicated security (?) parameter as well.

We already have it!? We can use Bearer token security scheme: https://superface.ai/docs/comlink/reference/provider#bearer-token

jnv:
> We already have it!? We can use Bearer token security scheme: https://superface.ai/docs/comlink/reference/provider#bearer-token

So, how do you pass the access token in the runtime into the `perform`?

PR #202, opened 2022-02-11
feat: Add `chat/members` with slack and discord

martinalbert:
## Description

Adds profile `chat/members` with use-case `GetMembers` and maps for this use-case.

## Types of changes

## Checklist:


grid/chat/members/profile.supr
```diff
@@ -0,0 +1,175 @@
...
+Get server members
+Retrieve paginated list of members
+"""
+usecase GetMembers {
+  input {
+    """
+    Server
+    Representation of a server. Can be ID or name, depending on the provider capability.
+    """
+    server string!
```
jnv, line 14: Server vs. workspace vs. ...

grid/chat/members/maps/discord.suma
```diff
@@ -0,0 +1,187 @@
...
+Operation MapMembers
+Returns array mapping members in received body
+"""
+operation MapMembers {
+  members = []
+
+  members = args.members.map(member => {
+    const resultMember = {
+      id: member.user.id,
+      joinedAt: Date.parse(member.joined_at),
```
jnv, line 60: Watch out, this will return a ms timestamp!

grid/chat/members/maps/slack.suma
```diff
@@ -0,0 +1,189 @@
...
+operation MapMembers {
+  members = []
+
+  members = args.members.map(member => {
+    const resultMember = {
+      id: member.id,
+      isBot: member.is_bot,
+    }
+
+    if (member.profile && member.profile.display_name) {
```
jnv, line 69: Doesn't Slack expose username's for internal purposes? Or it shouldn't be used?

jnv (changes requested):

grid/chat/members/maps/slack.suma
martinalbert: Since slack does not really describe which to use, I'm using `name` as a fallback. It seems [here](https://api.slack.com/changelog/2017-09-the-one-about-usernames) that they prefer to use `display_name` so I'm not sure.

martinalbert: Ok I found better description of user object and they do not recommend using `name` field ( see [here](https://api.slack.com/types/user) )

grid/chat/members/maps/discord.suma
martinalbert: Is that a problem?

jnv: Sure it isn't if you think that year `36681` is the same as `2004` (see `1095379198.75` vs `1095379198750`).

The documentation explicitly states UNIX timestamps giving me example in seconds, so...

grid/chat/members/maps/discord.suma
```diff
@@ -0,0 +1,187 @@
...
+
+  set if (!isNaN(resetTimestamp)) {
+    details.resetTimestamp = resetTimestamp * 1000
+  }
+
+  set if (!isNaN(retryAfter)) {
+    details.retryAfter = retryAfter
+  }
+
+  set if (details.remainingRequests && details.totalRequests) {
```
janhalama, line 117:
```suggestion
  set if (details.remainingRequests !== undefined && details.totalRequests) {
```

grid/chat/members/maps/slack.suma
```diff
@@ -0,0 +1,187 @@
...
+  set if (!isNaN(resetAfter)) {
+    details.resetAfter = resetAfter
+    details.resetTimestamp = Date.now() + (resetAfter * 1000)
+  }
+
+  set if (!isNaN(retryAfter)) {
+    details.retryAfter = retryAfter
+  }
+
+  set if (details.remainingRequests && details.totalRequests) {
```
janhalama, line 111:
```suggestion
  set if (details.remainingRequests !== undefined && details.totalRequests) {
```

janhalama (approved):
LGTM

grid/chat/members/profile.supr
```diff
@@ -0,0 +1,179 @@
...
+    
+    """
+    Limit
+    Maximum number of records to retrieve
+    """
+    limit number
+    
+    """
+    Page
+    Identification of page with paginated results (cursor)
```
jnv, line 28:
```suggestion
    Page cursor retrieved from nextPage for results pagination
```

grid/chat/members/profile.supr
```diff
@@ -0,0 +1,179 @@
...
+  joinedAt number
+
+  """
+  Is Bot
+  Boolean that represents whether the member is a bot
+  """
+  isBot boolean
+}
+
+model Members {
```
jnv, line 96: Nitpicking: I'd inline this model into use-case, or at least rename it to `MembersResult`.

grid/chat/members/maps/discord.suma
```diff
@@ -0,0 +1,187 @@
...
+        "Authorization" = `Bot ${parameters.accessToken}`
+      }
+    }
+
+    response 200 "application/json" {
+      members = call MapMembers(members = body)
+      rateLimit = call MapRateLimitDetails(headers = headers)
+
+      result.members = members,
+      result.nextPage = members[members.length - 1].id
```
jnv, line 29: This way I will either hit an error (if the array is empty), or the pagination will never end.

grid/chat/members/maps/discord.suma
```diff
@@ -0,0 +1,187 @@
...
+"""
+Operation MapMembers
+Returns array mapping members in received body
+"""
+operation MapMembers {
+  members = args.members.map(member => {
+    const resultMember = {
+      id: member.user.id,
+      joinedAt: Date.parse(member.joined_at),
+      isBot: member.user.bot ? true : false,
```
jnv, line 59: How about like this? :wink: 

 ```suggestion
      isBot: !!member.user.bot,
```

File: grid/chat/members/maps/discord.suma [hunk 10]
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    // map user's platform wide nickname - username
+    if (member.user && member.user.username) {
```
jnv, line 63: `member.user` condition doesn't make sense if you assume it exists on previous lines. I'd just move this into the construction of `resultMember` object.

grid/chat/members/maps/slack.suma
```diff
@@ -0,0 +1,187 @@
...
+    const resultMember = {
+      id: member.id,
+      isBot: member.is_bot,
+    }
+
+    if (member.profile && member.profile.display_name) {
+      resultMember.username = member.profile.display_name;
+    }
+
+    if (member.name && resultMember.username === undefined) {
```
jnv, line 71: So, technically, can I end up with empty username on Slack?

jnv (changes requested):
The biggest blocker for me is potentially faulty pagination on Discord.

grid/chat/members/profile.supr
```diff
@@ -0,0 +1,179 @@
...
+    Workspace
+    Representation of a workspace. Can be ID or name, depending on the provider capability.
+    """
+    workspace string!
+    
+    """
+    Limit
+    Maximum number of records to retrieve
+    """
+    limit number
```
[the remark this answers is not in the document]
martinalbert: Both providers are allowing this parameter, so why not?

grid/chat/members/profile.supr
martinalbert: This description is same across all chat use-cases 🤔

grid/chat/members/maps/discord.suma
martinalbert: Fair enough

grid/chat/members/maps/slack.suma
martinalbert: I wasn't sure if profile display name is not optional based on slack docs, but it's among common fields [here](https://api.slack.com/types/user), so I'll remove the checks.

grid/chat/members/maps/discord.suma
jnv: Do you want to include this suggestion?

jnv (changes requested):

grid/chat/members/maps/discord.suma
```diff
@@ -0,0 +1,191 @@
...
+      headers {
+        "Authorization" = `Bot ${parameters.accessToken}`
+      }
+    }
+
+    response 200 "application/json" {
+      members = call MapMembers(members = body)
+      rateLimit = call MapRateLimitDetails(headers = headers)
+
+      return map result if ((members.length < input.limit) || (members.length === 0)) {
```
jnv, line 28: This condition is technically correct, but confusing if `input.limit` is not set. Consider setting a default fallback value, perhaps something higher than Discord's default `1` (e.g. `100`).

grid/chat/members/profile.supr
martinalbert: I don't understand what's wrong with it 😄 what if user gets page cursor not from `nextPage`?

jnv: My problem is that this description is not very readable and understandable in the first place. I am not arguing where you get the cursor, although if you get it elsewhere, you probably know what you are doing and don't need this description.

grid/chat/members/maps/discord.suma
martinalbert: True, I'll have to fix it in other maps as well.

grid/chat/members/maps/discord.suma
```diff
@@ -25,14 +27,18 @@ map GetMembers {
       members = call MapMembers(members = body)
       rateLimit = call MapRateLimitDetails(headers = headers)
 
-      result.members = members,
-      result.nextPage = members[members.length - 1].id
-
-      set if (rateLimit) {
-        result.rateLimit = rateLimit
+      return map result if ((members.length < input.limit) || (members.length === 0)) {
```
jnv, line 30: This still uses `input.limit`, please check if there are possibly other issues like this.

jnv (changes requested):
Thanks for the fix, I see one more issue with the limit.

grid/chat/members/maps/discord.suma
martinalbert: I have checked other maps and only `chat/messages/discord` needs to be fixed. In this case, I assigned default limit value at the start of map.

jnv: Yes, but you don't use the limit. Or is using `input.limit` intentional here?

grid/chat/members/maps/discord.suma
```diff
@@ -0,0 +1,193 @@
...
+    title = "Input parameter `workspace` is required for provider discord"
+  }
+
+  limit = input.limit || 100;
+
+  http GET "/guilds/{input.workspace}/members" {
+    request "application/json" {
+      query {
+        after = input.page
+        limit = input.limit
```
jnv, line 18: Same here btw.

martinalbert:
Good catch 😅

jnv (approved):
Finally! :yay:

PR #213, opened 2022-02-22
feat(social-media/posts): Add LinkedIn map

jnv:
## Description

Add LinkedIn map for `social-media/posts` profile. Currently this doesn't expose `createdAt` because provider returns that as Unix timestamps and we have no easy way to convert those into ISO date (will be fixed once this feature is exposed in OneSDK).

I have opted out of shared tests to use a fixed profile ID and add test for pagination (this could be consolidated separately).

## Types of changes

## Checklist:


grid/social-media/posts/maps/linkedin.test.ts
```diff
@@ -0,0 +1,47 @@
...
+            useCase: 'GetProfilePosts',
+            input: {
+              profileId,
+            },
+          })
+        ).resolves.toMatchSnapshot();
+      });
+
+      it('correctly paginates', async () => {
+        for (let i = 0, page = null; i < 2; i++) {
```
janhalama, line 31: I found an error, when I try to paginate to third page the test starts to fail, because the page link query parameters returned by LinkedIn are uri encoded twice.

image

```
        for (let i = 0, page = null; i < 3; i++) {
```

grid/social-media/posts/maps/linkedin.suma
```diff
@@ -0,0 +1,107 @@
...
+        "X-Restli-Protocol-Version" = "2.0.0",
+        "Authorization" = `Bearer ${parameters.accessToken}`,
+      }
+    }
+  
+    response 200 "application/json" {
+      result = body.paging.links.reduce((acc, current) => {
+        const {rel} = current;
+        // Hack for "relativizing" URL
+        const href = current.href.substring(1);
```
janhalama, line 31: Just an idea: we could encode page references to be safe to use in query parameters.

janhalama (commented):
Overal looks good. LinkedIn API is hard to work with. See my comments.

grid/social-media/posts/maps/linkedin.suma
```diff
@@ -0,0 +1,107 @@
...
+            let videoUrl, duration, width, height;
+            if (attachmentType === 'video') {
+              const mediaElement = media['media~'] && media['media~'].elements.find((mediaEl) => {
+                return mediaEl.data['com.linkedin.digitalmedia.mediaartifact.AdaptiveStreamVideoWithAudio'] || mediaEl.data['com.linkedin.digitalmedia.mediaartifact.AdaptiveStreamVideo']
+              });
+              if (mediaElement) {
+                const artifact = mediaElement.data['com.linkedin.digitalmedia.mediaartifact.AdaptiveStreamVideoWithAudio'] || mediaElement.data['com.linkedin.digitalmedia.mediaartifact.AdaptiveStreamVideo']
+                width = artifact.videoStream.displaySize.width;
+                height = artifact.videoStream.displaySize.height;
+                duration = artifact.durationInMicroseconds / 1000;
```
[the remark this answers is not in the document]
jnv: Good point. So let's change it here like this?

```suggestion
                duration = artifact.durationInMicroseconds / (1000 * 1000);
```

grid/social-media/posts/maps/linkedin.suma
```diff
@@ -0,0 +1,107 @@
...
+            break;
+          case 'ARTICLE':
+            attachmentType = 'link';
+            break;
+          case 'NONE': // text-only posts; reshared posts?
+          case 'URN_REFERENCE': // polls?
+          case 'NATIVE_DOCUMENT': // document attachment (PDF)
+            break;
+        }
+        let attachments;
```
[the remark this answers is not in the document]
jnv: I am okay with this change _but_ we don't enforce this in the profile. Either we should enforce it, or (my preference) return undefined if no attachments are present. My preference is to avoid allocating empty arrays since attachments may be mostly empty with some providers.

grid/social-media/posts/maps/linkedin.test.ts
jnv: Good catch, I will check it.

grid/social-media/posts/maps/linkedin.suma
jnv: I have resorted to returning just the `start` parameter.

jnv:
@janhalama Thanks for the review, I have worked in your suggestions and expanded the tests a bit.

janhalama (approved):
Thanks for the fixes. LGTM 👍

PR #223, opened 2022-03-03
Add `ipstack` provider and `address/ip-geocoding` map

janhalama:
## Description

This PR adds `ipstack` provider and `address/ip-geocoding` map.

## Motivation and Context

We can not use `fields` query param to reduce the amount data being send, because time zone is not available for FREE subscription and if explicitly requested in `fields` ipstack API responds with error.

## Types of changes

## Checklist:


grid/address/ip-geolocation/maps/ipstack.test.ts
```diff
@@ -0,0 +1,30 @@
+import { RecordingDefinitions } from '@superfaceai/testing';
+
+import { ipGeolocationTest } from './ip-geocoding';
+
+ipGeolocationTest('ipstack', {
+  //Testing library is not processing recordings because integration parameter is used to replace HTTP scheme which is not supported by testing lib
+  processRecordings: false,
+  beforeRecordingSave: (recordings: RecordingDefinitions) => {
+    const IPSTACK_API_KEY_REDACTED_VALUE = 'IPSTACK_API_KEY_REDACTED';
```
jnv, line 9: This value should be better pulled out to the outer scope.

providers/ipstack.json
```diff
@@ -0,0 +1,24 @@
...
+      "id": "apikey",
+      "type": "apiKey",
+      "in": "query",
+      "name": "access_key"
+    }
+  ],
+  "parameters": [
+    {
+      "name": "HTTP_SCHEME",
+      "description": "HTTPS or HTTP scheme. The scheme identifies the protocol to be used to access the IP Stack API. Only HTTP scheme is allowed in case of IP Stack FREE subscription."
```
jnv, line 21: Wow, that's some weird monetization scheme. Would be useful to have explicit example (`"http" or "https"`) and mention what the default is.

Can we actually set the default?

grid/address/ip-geolocation/maps/ipstack.test.ts
janhalama: fixed

providers/ipstack.json
janhalama: I added explicit example and set the default to "https".

janhalama:
@jnv thanks for review I fixed all your comments

providers/ipstack.json
```diff
@@ -18,7 +18,8 @@
   "parameters": [
     {
       "name": "HTTP_SCHEME",
-      "description": "The scheme identifies the protocol to be used to access the IP Stack API. Allowed values are \"http\" or \"https\". Only \"http\" scheme is allowed in case of IP Stack FREE subscription."
+      "description": "The scheme identifies the protocol to be used to access the IP Stack API. Allowed values are \"http\" or \"https\". Only \"http\" scheme is allowed in case of IP Stack FREE subscription.",
+      "default": "https"
```
jnv, line 22: I am torn here about having `https` as default. `http` is supported by free subscription but it's unsafe by default. On one hand, you need to manually set this if you have free subscription, on the other hand if this was `http` by default, it would be unsafe from the start and it's possible paid subscribers wouldn't change it anyway.

So okay, keep `https` _however_ make sure there is sufficiently informative message if I attempt to use `https` while on the free subscription.

jnv (approved):
I have a concern regarding the `https` as a default scheme (just make sure it's properly reported), other than that it's good to go.

providers/ipstack.json
janhalama: The error returned by OneSDK communicates clearly what is the problem when using `https` while on the free subscription. See attached screenshot.

image

PR #230, opened 2022-03-23
feat(chat/messages): Implement retrieving thread messages

martinalbert:
## Description

Extends use-case `GetMessages` in profile `chat/messages` with new fields:
- `listThreadMessages`: boolean that represents whether to list messages from channel or thread destination.
- `threadId`: string that represents thread id

## Motivation and Context

Implement getting thread messages for slack.

## Types of changes

## Checklist:


jnv (changes requested):
TBH I think this approach is flawed, but maybe I misunderstand the context

- `listThreadMessages` seems redundant since it always needs to be set together with `threadId`
- But maybe it is needed and I can set `threadId` with `listThreadMessages = false`?
- Would it make sense to put this rather into a separate use case?

martinalbert:
Discord is able to get thread messages passing threadId as destination but slack has to specify thread id along with it which is not required in profile.

I can create new use-case for it which would remove `listThreadMessages` but discord use-case would be duplicate of its usecase `GetMessages` 🤷‍♂️

PR #250, opened 2022-05-05
feat: project management

tmladek:
## Description

Adds a minimal set of use cases for project management (creating a task and listing projects)

## Motivation and Context

Creating a task is an useful integration on its own, and the `ListProjects` usecase is needed so that we know what projects to create the task in.

## Types of changes

## Checklist:


grid/project-management/projects/profile.supr
```diff
@@ -0,0 +1,40 @@
+"""
+Manage projects in project management systems.
+"""
+name = "project-management/projects"
+version = "1.0.0"
+
+"""
+Create a new task.
```
jnv, line 8: Mind the description. Also, the first line is a title so it shouldn't be a sentence.

Since the use case doesn't have any input, it would be useful to mention how  the projects for listing are picked (are those all projects the user has permissions for?)

```suggestion
List Projects
```

grid/project-management/tasks/profile.supr
```diff
@@ -0,0 +1,71 @@
+"""
+Manage tasks in project management systems.
+"""
+name = "project-management/tasks"
+version = "1.0.0"
+
+"""
+Create a new task.
```
jnv, line 8:
```suggestion
Create a new task in project
```

grid/project-management/tasks/profile.supr
```diff
@@ -0,0 +1,71 @@
...
+    projects
+    An array of IDs of projects this task belongs to.
+    """
+    projects! [string]!
+
+		"""
+		label
+		Textual label of the task.
+		"""
+		label! string!
```
jnv, line 23: Is label Asana-specific lingo? I think I'd rather prefer `title` which seems more common?

grid/project-management/tasks/profile.supr
```diff
@@ -0,0 +1,71 @@
...
+Create a new task.
+"""
+usecase CreateTask {
+	input {
+
+    """
+    projects
+    An array of IDs of projects this task belongs to.
+    """
+    projects! [string]!
```
jnv, line 17: I think it would be safer to start with single project, ie.

```suggestion
    project! string!
```

...assuming most PM systems are rather hierarchical (just when I consider Trello or GitHub issues). Asana's multihoming is rather specific feature.

grid/project-management/tasks/profile.supr
```diff
@@ -0,0 +1,71 @@
...
+    """
+    assignee string
+	}
+
+	result {
+    "ID of the just created task."
+    id! string!
+
+    "Project IDs it was successfully added to."
+    projects! [string]!
```
jnv, line 49: This seems redundant?

grid/project-management/tasks/profile.supr
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+    assignee
+    ID of the user this task should be assigned to.
+    """
+    assignee string
+	}
+
+	result {
+    "ID of the just created task."
```
jnv, line 45:
```suggestion
    "ID of the created task."
```

grid/project-management/tasks/profile.supr
```diff
@@ -0,0 +1,71 @@
...
+    parent string
+
+    """
+    assignee
+    ID of the user this task should be assigned to.
+    """
+    assignee string
+	}
+
+	result {
```
jnv, line 44: I'd add something like `url` field with link to the created task.

providers/asana.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "asana",
+  "services": [
+    {
+      "id": "default",
+      "baseUrl": "https://app.asana.com/api/1.0"
+    }
+  ],
+  "securitySchemes": [
+    {
+      "id": "pat",
```
jnv, line 11:
```suggestion
      "id": "personalAccessToken",
```

jnv (changes requested):
Looks good, I have a few concerns about naming and making it too Asana-specific (like accepting array of projects from start)

grid/project-management/projects/profile.supr
tmladek: :facepalm: Copy & paste fail.

Agreed on the elaboration (though the use case will accept input in the future, alleviating the need for this a bit).

grid/project-management/tasks/profile.supr
tmladek: Nope, just my personal idiosyncrasy, I suppose. ("`label`/`description`" is a combination I feel like I often encounter).
FWIW, Asana calls it a `name` which didn't sit right with me.
I think `title` works fine though.

grid/project-management/tasks/profile.supr
tmladek: I thought that in that case, the map may just create multiple issues in different projects? Though I do see how the semantics differ now, though.

grid/project-management/tasks/profile.supr
tmladek: Not if the task failed to be added in some, but not other projects. This isn't the case for Asana which simply fails the whole operation, but it may be the case for other providers the maps of which would be creating multiple issues as described above. (If that's removed though, then yes, this is redundant.)

grid/project-management/tasks/profile.supr
```diff
@@ -5,7 +5,7 @@ name = "project-management/tasks"
 version = "1.0.0"
 
 """
-Create a new task.
+Create a new task in project
```
jnv, line 8:
```suggestion
Create a new task in a project
```

grid/project-management/tasks/profile.supr
```diff
@@ -14,7 +14,7 @@ usecase CreateTask {
     projects
     An array of IDs of projects this task belongs to.
```
jnv, line 15:
```suggestion
    project
    ID of project to create the task in.
```

grid/project-management/projects/maps/__snapshots__/asana.test.ts.snap
```diff
@@ -1,12 +1,14 @@
...
 exports[`project-management/projects/asana ListProject the usecase without any parameters should list all projects 1`] = `
-Ok {
-  "value": Array [
-    Object {
-      "id": "1202159846567214",
-      "name": "Customer implementation",
-    },
-  ],
+Err {
+  "error": "Security values for security scheme not found: personalAccesstoken
```
jnv, line 5: Watch out, this ended up in error, I asssume this is not intentional (and yeah, testing library doesn't expose error as a failure, it sucks, I know).

jnv (commented):
Looks good, consider reducing the strictness of fields (and in the doc string, I'd omit the ending dot `.` in the first line, which is rendered as name / title of the field).

Also be aware of errors in snapshots.

grid/project-management/tasks/profile.supr
```diff
@@ -17,10 +17,10 @@ usecase CreateTask {
     project! string!
 
 		"""
-		label
-		Textual label of the task.
+		title
+		Textual title of the task.
 		"""
-		label! string!
+		title! string!
```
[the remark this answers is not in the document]
tmladek: What is a task without a title? 🤔

grid/project-management/projects/maps/__snapshots__/asana.test.ts.snap
tmladek: 😬  it isn't

grid/project-management/tasks/profile.supr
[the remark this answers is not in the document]
jnv: An anonymous task ;)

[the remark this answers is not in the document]
tmladek: Interesting, Trello **UI** does everything in its might to stop you from having a card without a title... But the API accepts it just fine 🤷  
So I suppose the profile shouldn't require it either

grid/project-management/projects/maps/projects.ts
```diff
@@ -16,12 +16,13 @@ export const createProjectTest = (provider: string): void => {
...
-              useCase: 'ListProjects',
-              input: {},
-            })
-          ).resolves.toMatchSnapshot();
+          const result = await superface.run({
+            useCase: 'ListProjects',
+            input: {},
+          });
+
+          // console.log(result);
```
jnv, line 24: nit: console.log

grid/project-management/tasks/maps/tasks.ts
```diff
@@ -24,12 +24,13 @@ export const createTaskTest = (
...
-              useCase: 'CreateTask',
-              input: params,
-            })
-          ).resolves.toMatchSnapshot();
+          const result = await superface.run({
+            useCase: 'CreateTask',
+            input: params,
+          });
+
+          // console.log(result);
```
jnv, line 32: nit: console.log

jnv (approved):
LGTM, just forgotten console.logs

PR #262, opened 2022-05-27
chore: Rename mailchimp to mandrillapp

### LINKED ISSUE #248: [communication/send-(templated-)email] Confusion around Mailchimp × Mandrill
Currently both of our transactional profiles provide map for the `mailchimp` provider, which is confusing for users unfamiliar with Mailchimp's offering.

I just got from a support call with user who tried to use the Mailchimp map with Mailchimp's API key and was getting `Invalid key` error. The solution was to navigate to Transactional Email in Mailchimp's Dashboard and enter Mandrillapp dashboard.

Apparently the original intent behind the `mailchimp` provider was to respect that it's a single company and product. However, this seems to be confusing and could even turn problematic with other, Mailchimp-specific map.

My suggestion here is to introduce a new `mandrillapp` provider and change existing maps for Mailchimp to use that provider instead.

Before we do that, make sure that Mailchimp's APIs require a different API key for sure.

jnv:
## Description

Fixes #248.

This renames `mailchimp` provider to `mandrillapp` and changes both related maps (`communication/send-email` and `communication/send-templated-email`) to use `mandrillapp` instead.

I kept `mailchimp` app unchanged for the time being, it will have to be eventually changed to be able to target Mailchimp's API.

## Motivation and Context

This is usability enhancment, as we have found that current identification of "mailchimp" provider is confusing for users, because transactional emails are actually sent with "mandrillapp". This caused users to use Mailchimp's API keys and ending up with "Invalid key" errors.

## Types of changes

## Checklist:


freaz:
> I kept mailchimp app unchanged for the time being, it will have to be eventually changed to be able to target Mailchimp's API.

Can you elaborate what app do you mean?

freaz (approved):
The change itself looks fine. But I am wondering if you have a plan what will happen with `mailchimp` provider and maps already published to registry.

jnv:
> > I kept mailchimp app unchanged for the time being, it will have to be eventually changed to be able to target Mailchimp's API.
> 
> Can you elaborate what app do you mean?

Mailchimp provides [marketing API](https://mailchimp.com/developer/marketing/api/) at `https://*.api.mailchimp.com`; should we support that with other use cases in the future, we will have to modify the mailchimp provider definition.

> The change itself looks fine. But I am wondering if you have a plan what will happen with `mailchimp` provider and maps already published to registry.

It would be great if there were a possibility to hide the `mailchimp` provider on the relevant profiles' pages without removing the maps altogether, but I understand if it's not an option now. I assume the mailchimp maps will "disappear" with new minor versions of relevant profiles. I will also post an announcement on Discord and update our Resilient email tutorial.

freaz:
Oh ok, I got it now. We will need to add another service to support marketing APIs. 👍

freaz (approved):
Thanks for fixing it.

PR #279, opened 2022-08-17
feat: Support new testing library with sdk 2.1

martinalbert:
## Description
- Updates tools in `/src` to work with new testing library
- Updates snapshots, recordings and tests to work with new testing library

## Motivation and Context
New release of testing library with sdk 2.0

## Types of changes

## Checklist:


src/util.ts
```diff
@@ -14,21 +22,34 @@ export const EXTENSIONS = {
...
-let _superJson: SuperJson;
+let superJson: SuperJsonDocument;
 
-export function loadSuperJson(): SuperJson {
-  if (!_superJson) {
-    _superJson = SuperJson.loadSync().unwrap();
+export async function loadSuperJson(options?: {
+  fileSystem?: IFileSystem;
+}): Promise<{ document: SuperJsonDocument; path: string }> {
+  const superJsonPath = await detectSuperJson(
```
Jakub-Vacek, line 30: You run this even when `superJson` is set. It is not necessary - just check in `superJson` is defined, return it if is is or detect path and load if it isn't.

src/util.test.ts
```diff
@@ -42,11 +49,13 @@ describe('util', () => {
   });
 
   describe('loadSuperJson', () => {
-    it('should call loadSync once', () => {
-      const spy = jest.spyOn(SuperJson, 'loadSync');
+    it('should call loadSync once', async () => {
```
Jakub-Vacek, line 52: Is it still loadSync?

grid/address/clean-address/maps/clean-address.ts
```diff
@@ -10,11 +12,13 @@ export const cleanAddressTest = (
...
-        provider,
-        testInstance: expect,
-      });
+      superface = new SuperfaceTest(
+        {
+          profile: 'address/clean-address',
+          provider,
+        },
+        nockConfig
+      );
```
freaz, line 21: Silly question / nitpick: would be worth to create test helper containing factory function to create `SuperfaceTest` and add it to `globals` using https://jestjs.io/docs/configuration/#globals-object?

grid/test-config.ts
```diff
@@ -0,0 +1,5 @@
+import { NockConfig } from '@superfaceai/testing';
+
+export const nockConfig: NockConfig = {
+  testInstance: expect,
```
freaz, line 4: How was this done in previous version? How did we generate hashes for recordings?

src/util.ts
```diff
@@ -109,38 +157,61 @@ export function providersFiles(
...
+  }
 
-  return glob
-    .sync('../providers/*.json', { cwd })
-    .map(i => normalizePath(i, superJson));
+  return Promise.all(
+    glob
+      .sync('../providers/*.json', { cwd })
+      .map(async i => normalizePath(cwd, i))
+  );
```
freaz, line 176: suggestion: Shouldn't this be only:

```suggestion
  return glob
    .sync('../providers/*.json', { cwd })
    .map(i => normalizePath(cwd, i));
```

src/util.ts
```diff
@@ -109,38 +157,61 @@ export function providersFiles(
...
+  if (cwd === undefined) {
+    throw new Error('Super.json not found');
+  }
 
-  return glob
-    .sync('../grid/**/*.supr', {
-      cwd,
-    })
-    .map(i => normalizePath(i, superJson));
+  return Promise.all(
```
freaz, line 191: ditto

src/util.ts
```diff
@@ -109,38 +157,61 @@ export function providersFiles(
...
+  if (cwd === undefined) {
+    throw new Error('Super.json not found');
+  }
 
-  return glob
-    .sync('../grid/**/*.suma', {
-      cwd,
-    })
-    .map(i => normalizePath(i, superJson));
+  return Promise.all(
```
freaz, line 210: ditto

grid/address/clean-address/maps/clean-address.ts
martinalbert: It would definitely make sense now with `nockConfig` 👍🏻 I can try it on one test and let you know.

grid/test-config.ts
martinalbert: Mostly in same way (except those tests that did not use `testInstance`).`testInstance` was only moved from `profile`, `provider` and `useCase` parameters to second parameter in `SuperfaceTest` constructor.

src/util.ts
martinalbert: Yes, thanks.

grid/address/clean-address/maps/clean-address.ts
martinalbert: @freaz I had to add it with `setupFiles` since docs says that passing functions to `global` is only possible with `setupFiles` option, but typescript doesn't cooperate, see [this commit](https://github.com/superfaceai/station/pull/279/commits/e12442ac1512d2cc5f41125342406d7da4907507).

Jakub-Vacek (commented):
Looks ok, but first we need to release testing-lib in not beta version and use it here.

PR #281, opened 2022-09-07
feat: video upload

tmladek:
## Description

Adds a new profile for uploading videos via URL and a single provider map (Instagram)

## Types of changes

## Checklist:

- [?] All new and existing tests passed.


grid/social-media/publish-post/profile.supr
```diff
@@ -43,6 +43,10 @@ usecase PublishPost unsafe {
     media [
       MediaItem
     ]
+
+    attachments [
+      MediaContainer
```
jnv, line 48: I think MediaContainer is a term specific to Instagram, consider using naming the model like, for example, Attachment.
On the other hand, is it necessary to wrap the ID into an object? Since we need only an ID, consider changing this to just [string]. 

Also, you need to bump the minor version of the profile.

grid/social-media/upload-url/maps/upload-url.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export const publishVideoTest = (
+  provider: string,
+  name: string,
+  input: SuperfaceTestRun['input'],
+  hooks?: RecordingProcessOptions
+): void => {
+  describe(`social-media/upload-url/${provider}`, () => {
+    describe(`${name}`, () => {
```
jnv, line 33: Nit: you can use just name here, no need to wrap it into string interpolation.

grid/social-media/upload-url/profile.supr
```diff
@@ -0,0 +1,222 @@
+
+"""
+Upload Assets via URL to Social Media sites
+Upload images and videos available via an URL, onto social media such as Instagram, Facebook, or Twitter.
+"""
+
+name = "social-media/upload-url"
```
jnv, line 7: Nit: consider naming the profile like "upload-by-url" or "upload-from-url", we are not uploading the URL itself. ;)

grid/social-media/upload-url/profile.supr
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    error {
+      title = 'Bad request'
+    }
+  }
+}
+
+"""
+Get Asset State
```
jnv, line 105: Beside "media", "uploads", and "attachments", you are adding yet another term "asset" which increases confusion. Consider:

- renaming this to sth like GetUploadStatus
- changing attachments in publish-post to assets (or ecistingAssets) or vice versa

grid/social-media/upload-url/profile.supr
```diff
@@ -0,0 +1,222 @@
...
+    URL
+    URL of the asset. Must be publically acessible.
+    """
+    url! string!
+
+    """
+    Type
+    Media type of the asset
+    """
+    type enum {
```
jnv, line 38: This is risky: you are adding "reel" as enum, but this is specific just to Instagram. Al's type is very overloaded term. Consider changing this to a string and renaming it to provider Type. Alternatively make this optional and add new required field like "mimeType" from which we can infer a sane default.

jnv (changes requested):
I am on my phone so consider this as partial review.

grid/social-media/publish-post/profile.supr
tmladek: >I think MediaContainer is a term specific to Instagram, consider using naming the model like, for example, Attachment.

Attachment doesn't have the right connotations in my mind (but I'm open to changing it), is the issue outright *avoiding* provider-specific terms?

>On the other hand, is it necessary to wrap the ID into an object?

Pinterest, for ex. also includes a `cover_image_url` *at the moment of publishing*, and I can imagine other attributes of this kind, so I was trying to be future-proof

 >Also, you need to bump the minor version of the profile.

:+1:

grid/social-media/upload-url/profile.supr
tmladek: YouTube Shorts are pretty much 1:1 to Reels and doesn't map to "provider type" in the same way it does for Instagram

grid/social-media/upload-url/profile.supr
tmladek: true, a residue of about the 3rd renaming done :) will fix

grid/social-media/publish-post/profile.supr
jnv: > Attachment doesn't have the right connotations in my mind (but I'm open to changing it), is the issue outright avoiding provider-specific terms?

Yes. Also the name isn't related to the field and yet isn't used elswhere. Not sure what the correct term should be here. Maybe use "assets" for the field and "UploadedAssets" for the model to align this with the other profile?

> Pinterest, for ex. also includes a cover_image_url at the moment of publishing, and I can imagine other attributes of this kind, so I was trying to be future-proof

OK, make sense, though I am curious whether we can make an union out of it, like `[string | MediaContainer]`

grid/social-media/upload-url/profile.supr
tmladek: yes

jnv: The mapping is map's concern, my thought was that you'd provide input like `providerType: "short"` or `providerType: "reel"`. But maybe we could adopt some universal name like `short boolean`?

tmladek: not `short boolean` because that implies the video is *short* (whereas it also needs to be vertical), perhaps `reel boolean`, but then what about images?

grid/social-media/publish-post/profile.supr
tmladek: > Also, you need to bump the minor version of the profile.
Welp, should I bump versions of all maps then?

grid/social-media/upload-from-url/profile.supr
```diff
@@ -32,14 +32,19 @@ usecase RegisterUpload unsafe {
...
       image,
       video,
-      reel
     }
+
+    """
+    Short-form Video
+    Indicates a video should be uploaded as a short-form video (e.g. Instagram Reels)
+    """
+    shortForm boolean
```
jnv, line 47: Why not "shortFormVideo"?

grid/social-media/upload-from-url/profile.supr
```diff
@@ -32,14 +32,19 @@ usecase RegisterUpload unsafe {
     url! string!
 
     """
-    Type
+    Media Type
     Media type of the asset
     """
-    type enum {
+    mediaType enum {
```
jnv, line 38: Hmm, media type is another name for MIME type, so having this as enum can be somewhat misleading. But I don't have much better name for this (content type also smells of mime given the http header). How about mediaKind or contentKind?

jnv (commented):
Looks good, but please also rename the profile files for upload-from-url.

jnv (approved):
:shipit:

PR #282, opened 2022-09-20
feat: add video upload support for facebook

tmladek:
## Description

Adds support for Facebook video uploads

## Types of changes

## Checklist:


tmladek:
Known issues: `attachments` is at the moment always presumed to be a list of Video IDs, triggering special handling (updating the video with `published=true` instead of creating an actual post). There's nothing about the ID that could be used to tell whether it's a video or potentially a photo.

- add a `type` param to `MediaAttachment`?
- change `attachments` to `videos` (as that's the only use right now anyway?)
- `GET` various endpoints to see which one returns without error, giving info about the type of the item?
- ?

grid/social-media/publish-post/maps/facebook.suma
```diff
@@ -9,16 +9,22 @@ Publish post to Facebook profile (page)
...
     title = "profileId field is missing"
     detail = "Missing or invalid profileId, either page or group ID is expected"
   }
 
-  return map error if(media.length === 0 && !input.text && !input.link) {
-    title = "Missing text or link"
-    detail = "Either text or link fields must be non-empty for text posts"
+  return map error if (media.length === 0 && attachments.length === 0 && !input.text && !input.link) {
+    title = "Missing text, link or attachments"
+    detail = "Either text, link or attachments fields must be non-empty for text posts"
```
jnv, line 22: Nitpicking: this doesn't apply to just text posts, but _any_ posts

```suggestion
    detail = "Either text, link or attachments fields must be non-empty"
```

grid/social-media/publish-post/maps/facebook.suma
```diff
@@ -9,16 +9,22 @@ Publish post to Facebook profile (page)
...
+  }
+
+  return map error if(attachments.length > 1) {
+    title = "Only one attachment is supported by Facebook."
```
jnv, line 26: nit

```suggestion
    title = "Too many attachments"
```

grid/social-media/upload-from-url/maps/facebook.test.ts
```diff
@@ -0,0 +1,17 @@
+import { publishVideoTest } from './upload-from-url';
+
+publishVideoTest('facebook', 'single video', {
+  caption: 'Single video from Superface Station.',
+  url: 'https://download.samplelib.com/mp4/sample-5s.mp4',
+  uploadType: 'video',
+});
+
+/*
+publishVideoTest('facebook', 'single reel', {
```
jnv, line 10: Commented out code

grid/social-media/upload-from-url/maps/facebook.suma
```diff
@@ -0,0 +1,135 @@
...
+          case 'error':
+            return 'error';
+          case 'processing':
+            return 'inProgress'
+          default:
+            return null;
+        }
+      })();
+
+      map error if (state === null) {
```
jnv, line 59: I prefer having an object with mapped states over IIFE, but I understand that switch/case has more flexibility. For your consideraton:

```suggestion
      states = {
        ready: 'finished',
        error: 'error',
        processing: 'inProgress',
      }
      state = states[body.status.video_status]
      
      map error if (!state) {
```

jnv (commented):
LGTM, few nitpicks.

Regarding the attachments issue, I like this option from your suggestion:

> change attachments to videos (as that's the only use right now anyway?)

However, I am not sure how it will affect usage with Twitter (for images) and LinkedIn (for documents).

Technically that's a breaking change, but I don't think anyone's using that including ourselves? Correct order would be to keep attachments field, document it as deprecated with hint to use videos instead, and use attachments if it's set and videos isn't.
If we figure out we still need attachments in the future, we can "undeprecate" it.

The documentation for videos field should mention the upload-from-url profile.

jnv (approved):

grid/social-media/upload-from-url/maps/facebook.suma
tmladek: Good point, I think the `switch` statement is a leftover from when I had more logic in there, an object/map is cleaner.

tmladek: Well, except that the linter in `station` currently doesn't include the fix and so errors out :sweat_smile: The fix is already merged in `main` AFAIK, but upgrading the linter is beyond the scope of this PR I think

tmladek:
>However, I am not sure how it will affect usage with Twitter (for images) and LinkedIn (for documents).

Right - well, I suppose we'd have to add (yet) another attribute for that (`documents`, `images`)... So I'd rather vote for adding a `type` field?

jnv:
> Right - well, I suppose we'd have to add (yet) another attribute for that (documents, images)... So I'd rather vote for adding a type field?

That'd make it harder to use. I vote for `videos` with type of `[string]` and if we need to support more types, backward compat would be trivial.

tmladek:
Okay, one last thing I noticed that I want to point out before I change it - we already use `attachment`s with `type`s in station, even in the same category of profiles: https://github.com/superfaceai/station/blob/main/grid/social-media/posts/profile.supr#L78

jnv:
> Okay, one last thing I noticed that I want to point out before I change it - we already use `attachment`s with `type`s in station, even in the same category of profiles: https://github.com/superfaceai/station/blob/main/grid/social-media/posts/profile.supr#L78

That's good point, but I think it's a different use case (heh). That structure is for reading the data, here we optimize for writing / input. And you won't be feeding that structure from one profile to the other. In fact, having a similar structure under the same field name can be confusing.

PR #306, opened 2022-12-08
feat: Add `GetStageChanges` use case for `mock` provider

martinalbert:
## Description

Adds new profile `recruitment/get-stage-changes` with use case and map `GetStageChanges`

## Motivation and Context

ATS integrations

## Types of changes

## Checklist:


grid/recruitment/candidates/profile.supr
```diff
@@ -139,6 +140,108 @@ usecase CreateCandidate {
   }
 }
 
+"""
+Get Stage Changes
+Retrieves a list of stages for specific candidate throughout his recruitment process. The candidate does not have to be applied for a job.
+"""
+usecase GetStageChanges {
```
janhalama, line 147: suggestion: Move GetStageChanges into separate profile `recruitment/get-stage-changes`. There reasons are the a same as in case `CreateLead` use case. Having single use case per profile is the best approach for ATS domain harmonisation. Look at the ATS domain providers comparison table and supported use cases across all of them.

grid/recruitment/get-stage-changes/profile.supr
```diff
@@ -0,0 +1,198 @@
+"""
+Retrieve Stage Changes
+Lists stage changes for a candidate.
+"""
+name = "recruitment/get-stage-changes"
+version = "1.0.0"
+
+"""
+Get Stage Changes
+Retrieves a list of stages for specific candidate throughout his recruitment process. The candidate does not have to be applied for a job.
```
janhalama, line 10:
```suggestion
Retrieves a list of stages for specific candidate or lead throughout the recruitment process
```
We have `CreateCandidate` or `CreateLead` use cases. We should use the `Lead` term here to be consistent with the use cases.

grid/recruitment/get-stage-changes/profile.supr
```diff
@@ -0,0 +1,198 @@
...
+"""
+usecase GetStageChanges {
+  input {
+    """
+    Candidate Id
+    Unique identifier of a candidate.
+    """
+    candidateId! string!
```
janhalama, line 18: Can I use lead ID as well?

grid/recruitment/get-stage-changes/profile.supr
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  example Failed {
+    input {
+      candidateId = 'NOT_EXISTING_CANDIDATE_ID'
+    }
+
+    error {
+      title = 'Not found'
+      detail = 'Not found'
```
janhalama, line 69:
```suggestion
      detail = 'Candidate ID "NOT_EXISTING_CANDIDATE_ID" not found.'
```

grid/recruitment/get-stage-changes/profile.supr
martinalbert: This might be problematic with different providers. Depends on whether `Lead` and `Candidate` are different entities with separate id pools.

If they are not the same entity, it does not make sense to put lead ID here as `Lead` will always be just in `sourced` stage, but if they share the ids, it's possible and we could rename this field.

janhalama: Our goal is to harmonise the APIs so that the client using OneSDK doesn't have to care about the implementation detail on the provider side.

50inTech is very specific about this use case:
> Fetching the status changes of a previously sent candidate (applicant or sourced)

We should accept both Candidate ID and Lead ID.

In case that Lead and Candidate are different entities with separate ID pools on the provider side. Can we add ID of the pool to Candidate or Lead entity ID when retrieving it using `ListCandidates`, `CreateCandidate or `CreateLead` use cases?

martinalbert: Good point. I'm not sure if I understand the question, do you mean to alter the id returned from those use cases and parse it in this use case to differ between candidate/lead?

janhalama: Proposed solution: Use cases `ListCandidates`, `CreateCandidate` or `CreateLead` return compound ID for example: "CANDIDATE:1" or "LEAD:10". Use case `GetStageChanges` parses the ID and handles returns result differently for lead IDs and candidate IDs.

martinalbert: 👍🏻

martinalbert: Updated documentation, didn't find better naming then `candidateId` 🤷🏻‍♂️

janhalama (approved):
LGTM

PR #307, opened 2022-12-09
feat: Add `ai/generate-text` with 2 use cases

kysely:
## Description

Adding OpenAI's GPT-3 text capabilities!

`ai/generate-text`
- `CompleteText`: Generate a text completion that attempts to match whatever context or pattern you give it
- `EditText`: Edit a text using the given instructions

Provided abstractions for tokens (what are those, people think in words and characters) and temperature (which really is how creative the model should be). Also used simpler model sizes.

## Questions

1. ~Is `gpt` scope correct? We already have [`language`](https://superface.ai/catalog#language) but it feels a bit different. What about `ai` scope? I also think about adding use case for generating images based on the prompt~ — using `ai` scope

## Preview

Screenshot 2022-12-09 at 16 04 25

Screenshot 2022-12-09 at 16 04 33

## Types of changes

## Checklist:


jnv:
> Is gpt scope correct? We already have language but it feels a bit different. What about ai scope? I also think about adding use case for generating images based on the prompt

How about "Natural Language Processing" or `nlp`? I think that'd work also for image generators.

janhalama:
> Is gpt scope correct? We already have [language](https://superface.ai/catalog#language) but it feels a bit different. What about ai scope?

I like `ai` scope more than `gpt`. GPT stands for Generative Pre-trained Transformer which describes how the AI works under the hood. AI is more generic.

grid/gpt/generate-text/maps/openai.suma
```diff
@@ -0,0 +1,180 @@
...
+  // Clamp `creativity` between 0 and 1
+  return Math.min(Math.max(creativity, 0), 1)
+}
+
+operation MapMaxLength {
+  // About tokens: https://beta.openai.com/tokenizer
+
+  DEFAULT_MAX_WORDS = 12
+
+  return parseInt((args.approxMaxWords || DEFAULT_MAX_WORDS) * 1.4)
```
janhalama, line 112: question: So you are transforming words to OpenAI tokens. The 1.4 constant will work for English, what about other languages?

grid/gpt/generate-text/maps/openai.suma
```diff
@@ -0,0 +1,180 @@
...
+    title = "Unauthenticated"
+    detail = "Please make sure you're providing a valid OpenAI API key"
+  }
+
+  return if (statusCode === 403) {
+    title = "Not allowed"
+  }
+
+  return if (statusCode === 429) {
+    title = "Quota limit exceeded"
```
janhalama, line 168: suggestion: I do not think that title should be used by the client to handle the error type. Using explicit enum of error types that the client can expect to be returned by the use case is clearly defined contract of the profile.

janhalama (commented):
I like the way you have written the profile and map, for example by mapping words to tokens the profile is prepared for new providers.

grid/gpt/generate-text/maps/openai.suma
kysely: > The 1.4 constant will work for English, what about other languages?

For English, the actual best ratio is somewhere around `1.2`. I tried Czech phrases and it was around `1.6`, so I thought 1.4 might be fine. For that reason, I also call the property in profile an _approximate_.

Do you find this a blocker for using?

janhalama: > Do you find this a blocker for using?

IMO for basic use cases this abstraction will be good enough.

On the other hand tokens are core of the GPT-3 technology. OpenAI monetisation is based on them. There is this OpenAI document describing how complex the topic is https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them. We are hiding the complexity by using inaccurate constant.

grid/gpt/generate-text/maps/openai.suma
kysely: Good point. I see this is becoming a new convention and I like it 👍 Added the code

grid/gpt/generate-text/maps/openai.suma
kysely: The complexity is actually the reason for abstraction, though, no?

My thinking is:
— I don't know how tokens are calculated
— They are different for every language (which means I need to use different values for different languages anyway)
— I kind of still need to experiment with it to determine the optimal max tokens value

So to me, it feels like it's _“random”_ anyway, but the word abstraction makes it easier for me to at least get some idea (approximate) of what I should be expecting.

Good point on the monetization, though. However don't the points above apply as well?

kysely: BTW Please don't take this as arguing your points. I'm trying to figure what's the best for the profile. Leaking the _tokens_ thing in there feels 🤢 but if it would be totally necessary, then we'll do it 🤷

janhalama: > The complexity is actually the reason for abstraction, though, no?
I agree.

janhalama (approved):
LGTM 👍

grid/gpt/generate-text/maps/openai.suma
kysely: I see now. Thank you 🙏

PR #311, opened 2022-12-14
feat: Add `GetStageChanges` map for provider `breezy-hr`

martinalbert:
Adds new map `GetStageChanges` for provider `breezy-hr`.

## Description
ATS integrations

## Types of changes

## Checklist:


grid/recruitment/get-stage-changes/maps/__snapshots__/breezy-hr.test.ts.snap
```diff
@@ -0,0 +1,63 @@
...
+exports[`recruitment/get-stage-changes/breezy-hr GetStageChanges when specified candidate does not exist returns error 1`] = `
+Err {
+  "error": "MappedHTTPError: Expected HTTP error
+Properties: {
+  \\"title\\": \\"Unknown error\\",
+  \\"detail\\": {
+    \\"type\\": \\"getStreamError\\",
+    \\"message\\": \\"there was an unknown error getting candidate stream.\\"
+  },
+  \\"code\\": \\"UnknownError\\",
```
janhalama, line 51: This is not good that the API returns 500 error in this case. I am afraid we are not able to solve it as they return 500 anytime they get wrong path params.

grid/recruitment/get-stage-changes/maps/breezy-hr.suma
```diff
@@ -0,0 +1,133 @@
+profile = "recruitment/get-stage-changes@1.0"
+provider = "breezy-hr"
+
+map GetStageChanges {
+  // TODO: add this error code to profile or use 'BadRequest'?
```
janhalama, line 5: I would go with `BadRequest` error, because this is limitation of the provider. The client needs to know that JobId is required input before perform of the use case. We should provide the information somewhere. It is a good question for the authoring feedback.

martinalbert: True 👍🏻

grid/recruitment/get-stage-changes/maps/__snapshots__/breezy-hr.test.ts.snap
martinalbert: Yeah, no way to show some helpful info here 😞

janhalama (approved):
LGTM

grid/recruitment/get-stage-changes/maps/__snapshots__/breezy-hr.test.ts.snap
```diff
@@ -0,0 +1,63 @@
...
+      Object {
+        "createdAt": "2022-12-14T11:23:18.812Z",
+        "current": true,
+        "id": "applied",
+        "name": "Applied",
+      },
+      Object {
+        "createdAt": "2022-12-14T11:21:30.696Z",
+        "current": false,
+        "id": "feedback",
```
janhalama, line 16: It is strange that the ID of the stage is not unique in the collection.

martinalbert: There are also unique identifiers, but since it's pulling it from `stream` endpoint, there are also other events and those ids are tied to events, not stages.

martinalbert: I could add model `StageChange` and have id of a event and additional field `stageId` which does not have to be unque, what do you think?

janhalama: > I could add model StageChange and have id of a event and additional field stageId which does not have to be unique, what do you think?

I do not get it, can you elaborate it more.

The option here could be to filter feedback out, as it is not recruitment stage.

martinalbert: the endpoint offers events and contain unique ids for those events, but stage ids can be duplicated in these events as I do not request for list of stages, but stage changes (candidate stream). I propose to change the current model `Stage` in profile to model `StageChange` to mimic more an event entity rather than stage entity. We can add model `Stage` for use case to list available stages.

martinalbert: or maybe `description` does not have to be dropped as it can describe reason of stage change

janhalama: Now I understand. It is actually valid for the candidate to be in the same stage multiple times.

janhalama: OK lets make the profile changes you described `Stage` -> `StageChange` and `id` -> `stageId`

janhalama: @martinalbert lets merge this PR and do the Stage -> StageChange and id -> stageId change in new PR

janhalama (approved):
Lets merge this PR and do the change discussed here in subsequent PR.

PR #313, opened 2022-12-15
feat: Add `CreateLead` use case for provider `breezy-hr`

martinalbert:
## Description
Adds new map `CreateLead` for provider `breezy-hr`.

## Motivation and Context
ATS integrations.

## Types of changes

## Checklist:


grid/recruitment/create-lead/maps/__snapshots__/breezy-hr.test.ts.snap
```diff
@@ -0,0 +1,40 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`recruitment/create-lead/breezy-hr CreateLead should map error 1`] = `
+Err {
+  "error": "MappedError: Expected error
+Properties: {
+  \\"title\\": \\"Unknown error\\",
```
janhalama, line 7: I still do not like that we return unknown error in case that JobId doesn't exist. I suggest this job ID validation:

```
 operation CheckJobExists {
      http GET "/v3/company/{ args.companyId }/position/{ args.jobId }" {

      request {
        headers {
          Authorization = args.accessToken
        }
      }

      response 200 {       
        return {}
      }

      response 500 "application/json" {
        //Breezy HR returns http status code 500 in case that the job ID doesn't exist. It could be caused by other reasons as well, but at least the error helps the client understand the error better.
        fail {
          title = 'Bad Request'
          detail = 'Can not get job ID ' + args.jobId + '. Check that the job ID exists.'
          code = 'BadRequest'
          rateLimit = call MapRateLimit(headers = headers)
        }
      }

      response "application/json" {
        error = call MapBreezyHRError(
          statusCode = statusCode,
          error = body.error,
          headers = headers
        )

        fail error
      }

      response {
        fail {
          title = "Unknown error"
          detail = JSON.stringify(body, null, 2)
          code = 'UnknownError'
        }
      }
    }
  }
```

janhalama: I do not like using`BadRequest` code here. HTTP `BadRequest` errors indicates problem in the map. `NotFound` is actually better code here and consistent with `workable` provider.

Another topic for authoring feedback.

janhalama (approved):
LGTM

PR #317, opened 2022-12-16
feat: Include entity type in Id for use case `CreateCandidate`

martinalbert:
## Description
Adds entity type prefix to id of candidate.

## Motivation and Context
Discussion in [this comment](https://github.com/superfaceai/station/pull/306#discussion_r1049570684), main reason is support other use cases with different providers.

## Types of changes

## Checklist:

### DIFF
--- a/grid/recruitment/create-candidate/maps/__snapshots__/breezy-hr.test.ts.snap
+++ b/grid/recruitment/create-candidate/maps/__snapshots__/breezy-hr.test.ts.snap
@@ -3,7 +3,7 @@
 exports[`recruitment/create-candidate/breezy-hr CreateCandidate when specified job does exist performs correctly 1`] = `
 Ok {
   "value": Object {
-    "id": "b23ca444e77801",
+    "id": "CANDIDATE:b23ca444e77801",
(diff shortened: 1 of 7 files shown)


janhalama:
@martinalbert do we need it in case of Breezy HR provider? If I remember it well Breezy HR API has single id pool for candidates and leads.

martinalbert:
Yes, workable should also have common pool, this was meant for other providers.

janhalama:
> Yes, workable should also have common pool, this was meant for other providers.

I do not see the reason for adding the entity prefix in case of providers with single id pool for both candidates and leads. Structure of the ID is opaque for the client. We should be able to use directly the ID of the entity in case of Breezy HR and Workable providers.

martinalbert:
Ok I misunderstood, it's different with breezy, but both have common id pool for lead and candidate. Breezy does not have stage `sourced`, you create a candidate with origin of `sourced`, but starts in first stage of a pipeline which is by default `applied`.

martinalbert:
Closing this - not needed at the moment.

PR #324, opened 2022-12-21
DRAFT: Update one-sdk to v2.2.0

janhalama:
## Description

This PR updates `@superfaceai/one-sdk` to version v2.2.0.

This is DRAFT PR and must not be merged into main until stable version of v2.2.0 is released.

## Motivation and Context

We need update version of OneSDK with binary support to upload CVs in  ATS vertical.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I haven't repeated the code. (DRY)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #326, opened 2022-12-28
`recruitment/create-candidate` 2.0 profile

janhalama:
## Description

This PR adds `recruitment/create-candidate` 2.0 profile version with following changes:
- `cv` input fields changed from { data: string, name: string } to { data, fileName: string }, type of data is not specified, but it accepts `BinaryData` and `name` field has been changed to `fileName` as it better describes its purpose.
- new use case `CreateCandidateFeatures` added with list of features supported by the provider
- `CVFileNameRequired` and `CVMIMETypeNotSupported` error types added

Based on #324 

## Motivation and Context

Support CV upload for `breezy-hr` provider.

## Types of changes

## Checklist:


grid/recruitment/create-candidate/profile.supr
```diff
@@ -139,6 +139,45 @@ usecase CreateCandidate {
...
+      The cv input contains url field with publicly accessible URL of a CV file.
+      """
+      url = 'url'
+      """
+      File
+      The cv input contains data field with CV file and fileName field with the file name.
+      """
+      file = 'file'
+    }!
+  ]!
```
martinalbert, line 170:
```suggestion
    """
    CV MIME Types
    List of CV MIME types accepted by the provider. MIME types are defined in RFC2045 and registered by https://www.iana.org/.
    """
    cvMIMETypes! [string!]!
    """
    CV Upload Methods
    List of supported CV upload methods.
    """
    cvUploadMethods! [
      enum {
        """
        URL
        The cv input contains url field with publicly accessible URL of a CV file.
        """
        url = 'url'
        """
        File
        The cv input contains data field with CV file and fileName field with the file name.
        """
        file = 'file'
      }!
    ]!
```

grid/recruitment/create-candidate/maps/breezy-hr.suma
```diff
@@ -242,3 +264,66 @@ operation CheckCandidateByEmail {
...
+operation UploadCV {
+  cv = args.candidate.cv || {}
+
+  return if(cv.data === undefined || cv.data === null) {}
+
+  tmp = (cv.data.name = cv.fileName)
+
+  http POST "/v3/company/{ args.companyId }/position/{ args.candidate.jobId }/candidate/{ args.candidateId }/resume" {
+
+  request "multipart/form-data" {
```
martinalbert, line 297:
```suggestion
    request "multipart/form-data" {
```

grid/recruitment/create-candidate/maps/breezy-hr.test.ts
```diff
@@ -1,6 +1,53 @@
...
+          //resume upload contains multipart/form-data body with boundary with randomized value, we replace the boundary with constant in recorded body and matched body
+          const recordedBody = replaceBoundaryInMultipartFormDataBody(
+            boundaryReplacement,
+            recording.body
+          );
+          recording.body = body => {
+            const requestBody = replaceBoundaryInMultipartFormDataBody(
+              boundaryReplacement,
+              body
+            );
```
martinalbert, line 31: question: why is this needed? 🤔

grid/recruitment/create-candidate/maps/recordings/breezy-hr.recording.json
```diff
@@ -255,16 +280,237 @@
...
       }
+    ],
+    "0de0fe70126c707d7921a8307e36c0ec": [],
+    "29a5bf03ceca5dbd3e7d244d42b01174": [],
+    "4d58e4b9dbfa220a6ddad83fe9d25e7e": [],
+    "8d1596557a6ba534defec4d4d9016532": [],
+    "cfa81861f9e39455b6c83fc15d1942da": [],
+    "f4dc24e7230ff97506452f9a9ecaa2c2": [],
+    "1f557bdc0faa8ff1c45d6772a897d773": [
+      {
```
martinalbert, line 301: Aren't these duplicated? (see also bellow)

grid/recruitment/create-candidate/maps/recordings/workable.recording.json
```diff
@@ -249,10 +249,212 @@
...
+        ],
+        "responseIsBinary": false
+      }
+    ],
+    "d92c273026c27383eaa78974b272b2ff": [],
+    "1a1c1f0b9161bd1f6848dbafdeec6e46": [],
+    "f5d3ca6e0319573833951afff46d2008": [],
+    "e14b0dd6837efbd20652e1a90b21ceef": [],
+    "323c2ff09bf66f9d2eb60ab73e0d2095": [
+      {
```
martinalbert, line 262: Also seems like duplication.

grid/recruitment/create-candidate/maps/breezy-hr.test.ts
janhalama: The boundary needs to be replaced with constant value in recorded and tested requests because boundary changes with each subsequent request even when the rest of the request contains the same data.

This will be the case of every multipart/form-data request tested using nock and @superfaceai/testing lib. We should come with generic solution in @superfaceai/testing library.

grid/recruitment/create-candidate/maps/breezy-hr.test.ts
```diff
@@ -1,6 +1,53 @@
...
+    beforeRecordingLoad: recordings => {
+      recordings.forEach(recording => {
+        if (
+          recording.method === 'POST' &&
+          /^\/v3\/company\/.+\/position\/.+\/candidate\/.+\/resume$/m.test(
+            recording.path
+          )
+        ) {
+          const boundaryReplacement = '---boundary replaced';
+          //resume upload contains multipart/form-data body with boundary with randomized value, we replace the boundary with constant in recorded body and matched body
```
janhalama, line 22:
```suggestion
          //Requests with multipart/form-data content type contains boundary in request body. Boundary value changes even when the rest of the request contains the same data.
```

grid/recruitment/create-candidate/maps/breezy-hr.test.ts
janhalama: See my suggestion with better description in comment above.

grid/recruitment/create-candidate/maps/recordings/workable.recording.json
janhalama: fixed

grid/recruitment/create-candidate/maps/recordings/breezy-hr.recording.json
janhalama: Fixed, but I am not sure what is the right testing flow to avoid duplications. Looks like I have to delete file with recording before calling 
```
yarn test:record grid/recruitment/create-candidate/maps/breezy-hr -u
```
in order to avoid duplicate snapshots.

martinalbert: Currently the best way is to always try the test with mocked traffic and either:
- fix a hash for an old recording to one that can't be found (Mentioned in error message)
- or remove recordings that you think are duplicated and rerecord

grid/recruitment/create-candidate/maps/breezy-hr.test.ts
martinalbert: I understand and agree that this should be added in processing of recordings logic in testing library. I have two more questions:
- why do you assign function to a body?
- isn't there a need for `beforeRecordingSave` implementation? Does it work only with this (record -> test with recorded traffic)?

janhalama: > why do you assign function to a body?

The function is custom body matcher implementation. It is supported by nock library.

> isn't there a need for beforeRecordingSave implementation

Well we could change the body before we save the recording, but that body would still not match the body generated during the test run. Boundary is random value.

martinalbert: Thanks for clarifying, was just wondering if we don't want to replace boundary in `beforeRecordingSave` and keep recording without this dynamic boundary, my only concern is that recording body will change each time we rerecord it and it is not clear what changed in hex format. 🤷🏻‍♂️

martinalbert (approved):
LGTM

grid/recruitment/create-candidate/maps/breezy-hr.test.ts
janhalama: > just wondering if we don't want to replace boundary in beforeRecordingSave and keep recording without this dynamic boundary

It is the request body that is changing, not response body. Keep recording without dynamic boundary would not help, because the request body boundary generated during test run would not match saved recording.

martinalbert: Wouldn't it match recordings with your implementation of `beforeRecordingLoad` (with use of function as a body matcher)?

PR #341, opened 2023-01-17
feat: add CRM list profiles (get lists, add contact to list)

tmladek:
## Description

Adds `GetLists` and `AddContactsToList` use cases.

## Types of changes

## Checklist:

### DIFF
--- a/grid/crm/lists/maps/sendgrid.suma
+++ b/grid/crm/lists/maps/sendgrid.suma
@@ -1,0 +1,52 @@
+profile = "crm/lists@1.0"
+provider = "sendgrid"
+
+map GetLists {
+    http GET "/v3/marketing/lists" {
+        security "bearer_token"
+
+        request {
+            query {
+                page_token = input.page,
+            }
+        }
+
+        response 200 "application/json" {
+            map result {
+                lists = body.result.map((list) => {
+                    return {
+                        listId: list.id,
+                        name: list.name,
+                        length: list.contact_count
+                    }
+                })
+                nextPage = body._metadata.next ? body._metadata.next.match('page_token=([a-z0-9-]+)')[1] : null
+            }
+        }
+    } 
+}
+
+// If the contact already exists in the system, any entries submitted via this endpoint
+// will update the existing contact. The contact to update will be determined only by the
+// email field and any fields omitted from the request will remain as they were. A 
+// contact's ID cannot be used to update the contact.
+//
+// https://docs.sendgrid.com/api-reference/contacts/add-or-update-a-contact
+map AddContactsToList {
+  http PUT "/v3/marketing/contacts" {
+    security "bearer_token"
+
+    request {
+      body = {
+        list_ids: [input.listId],
+        contacts: input.emails.map((email) => {
+            return {
+                email: email
+            }
+        })
+      }
+    }
+
+    response 202 {}
+  }
+}
--- a/grid/crm/lists/profile.supr
+++ b/grid/crm/lists/profile.supr
@@ -1,0 +1,96 @@
+"""
+Lists
+Manage contact lists in Customer Relation Management systems.
+"""
+
+name = "crm/lists"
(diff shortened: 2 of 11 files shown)


grid/crm/lists/maps/sendgrid.suma
```diff
@@ -0,0 +1,52 @@
...
+                        length: list.contact_count
+                    }
+                })
+                nextPage = body._metadata.next ? body._metadata.next.match('page_token=([a-z0-9-]+)')[1] : null
+            }
+        }
+    } 
+}
+
+// If the contact already exists in the system, any entries submitted via this endpoint
```
janhalama, line 29: suggestion: I do not think that it is right that the map implements`upsert` but the name of the use case is `AddContactsToList`. I suggest to change use case name to `AddOrUpdateContacts`.

I checked that MailChimp provider which supports `upsert` as well.

grid/crm/lists/maps/sendgrid.suma
```diff
@@ -0,0 +1,52 @@
...
+        list_ids: [input.listId],
+        contacts: input.emails.map((email) => {
+            return {
+                email: email
+            }
+        })
+      }
+    }
+
+    response 202 {}
```
janhalama, line 50: suggestion: handle error responses

grid/crm/lists/profile.supr
```diff
@@ -0,0 +1,96 @@
+"""
+Lists
+Manage contact lists in Customer Relation Management systems.
+"""
+
+name = "crm/lists"
```
janhalama, line 6: question: What are the benefits of single `crm/lists` profile compared to `crm/get-lists` and `crm/add-contacts-to-lists` profiles ?

 I am not big fan of profiles build around entities. It is hard to harmonise such profiles. All providers needs to implement all of the use cases inside the profile. Comlink by its nature gives us the option to not implement profile by provider.

grid/crm/lists/maps/sendgrid.suma
tmladek: The use case is "add a contact to a contact list", and the map implements that; for this provider, it's using an endpoint which has upsert semantics, but there's actually no updating going on.

grid/crm/lists/profile.supr
tmladek: It's a more complete (application) profile - it's hard to add to a list whose ID I cannot obtain. 

But that's a deeper and more general issue, if the convention is to go with per-use case profiles, let's do that for now

grid/crm/lists/maps/sendgrid.suma
janhalama: OK, i understand. There is still one disproportion that should be harmonised across providers.  AddContactsToList can either:
- do nothing if already existing email is being added (which is this case)
- fail if already existing email is being added

We should have test for it.

grid/crm/lists/profile.supr
janhalama: > it's hard to add to a list whose ID I cannot obtain

I understand, you are saying that it doesn't make sense to have one use without the other.

janhalama: Let's keep it then. I do not know CRM domain properly and do not know the demands.

grid/crm/add-contacts-to-list/profile.supr
```diff
@@ -0,0 +1,43 @@
...
+
+"""
+Add Contacts to List
+Adds contacts to an existing contact list.
+"""
+usecase AddContactsToList unsafe {
+  input {
+    """
+    List ID
+    Unique ID of the list to add a contact to.
```
janhalama, line 17:
```suggestion
    Unique ID of the list to add a contact to. Use `crm/lists` profile to get the ID.
```

janhalama (approved):
LGTM 👍

I have one minor suggestion to improve profile docs.

PR #349, opened 2023-01-30
feat: add ListJobs map for greenhouse

tmladek:
## Types of changes

## Checklist:

### DIFF
--- a/grid/recruitment/list-jobs/maps/greenhouse.suma
+++ b/grid/recruitment/list-jobs/maps/greenhouse.suma
@@ -18,6 +18,7 @@ map ListJobs {
 
   page = input.page || 1
   limit = input.limit || 100
+  status = input.state && JOB_STATES_S2G[input.state]
   
   http GET "/v1/jobs" {
     security "apiKeyBasicAuth"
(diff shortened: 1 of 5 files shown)


grid/recruitment/list-jobs/maps/greenhouse.suma
```diff
@@ -0,0 +1,60 @@
...
+  limit = input.limit || 100
+  
+  http GET "/v1/jobs" {
+    security "apiKeyBasicAuth"
+  
+    request {
+      query {
+        page = page
+        per_page = limit
+        status = JOB_STATES_S2G[input.state]
```
jnv, line 29: issue: Since state is not mandatory, I may end up passing `JOB_STATES_S2G[undefined]` here.

jnv (changes requested):
Looks good, I see one critical issue.

grid/recruitment/list-jobs/maps/greenhouse.suma
tmladek: In which case it evaluates to `undefined` and isn't passed in the `query` at all, right?

jnv: Right, it should, I just don't like that it relies on implicit behavior. Feel free to ignore this suggestion then.

```suggestion
        status = input.state && JOB_STATES_S2G[input.state]
```

(even better would be to assign state above the http call)

jnv (approved):

PR #350, opened 2023-01-30
Prepare for OneSDK 2.3.0

freaz:
## Description

- updates Parser to 2.1.0
- updates AST to 1.3.0
- updates OneSDK to 2.3.0
- updates Testing to 4.0.0
-  prepares maps which could be failing with new OneSDK 2.3.0

Requires https://github.com/superfaceai/station/pull/354

## Types of changes

## Checklist:

### DIFF
--- a/grid/address/geocoding/maps/__snapshots__/azure-maps.test.ts.snap
+++ b/grid/address/geocoding/maps/__snapshots__/azure-maps.test.ts.snap
@@ -38,12 +38,10 @@ Ok {
 
 exports[`address/geocoding/azure-maps ReverseGeocode when inputs are invalid should throw an exception 1`] = `
 Err {
-  "error": "MappedHTTPError: Expected HTTP error
-Properties: {
-  \\"title\\": \\"Bad request\\",
-  \\"detail\\": \\"The provided coordinates in query are invalid, out of range, or not in the expected format\\"
-}
-AST Path: definitions[1].statements[1].responseHandlers[1].statements[0]
-Original Map Location: Line 130, column 7",
+  "error": "InputValidationError: Input validation failed:
+Path: input.latitude
+Error: Missing required field
+Path: input.longitude
+Error: Missing required field",
 }
 `;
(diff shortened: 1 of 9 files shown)


grid/address/geocoding/maps/__snapshots__/azure-maps.test.ts.snap
```diff
@@ -38,12 +38,10 @@ Ok {
...
 exports[`address/geocoding/azure-maps ReverseGeocode when inputs are invalid should throw an exception 1`] = `
 Err {
-  "error": "MappedHTTPError: Expected HTTP error
-Properties: {
-  \\"title\\": \\"Bad request\\",
-  \\"detail\\": \\"The provided coordinates in query are invalid, out of range, or not in the expected format\\"
-}
-AST Path: definitions[1].statements[1].responseHandlers[1].statements[0]
-Original Map Location: Line 130, column 7",
+  "error": "InputValidationError: Input validation failed:
```
freaz, line 41: Here we could also send input as `{ latitude: undefined, longitude: undeifned }` as we would get the same result as before.

Which would pass [required field](https://superface.ai/docs/comlink/profile#sec-Required-field) validation and thus the request would be send.

grid/chat/members/maps/members.ts
```diff
@@ -21,12 +21,17 @@ export const getMembersTest = (
       });
 
       it('performs correctly', async () => {
+        const input: Record<string, number | string> = {
+          limit: 1,
+        };
+
+        if (typeof options?.workspace === 'string') {
+          input.workspace = options?.workspace;
```
freaz, line 29: workspace in profile is defined as `workspace: string!`, which means if the field exists it must be a string. Here it can be `undefined` which is invalid. So, the `workspace` should be set only if it is a `string`.

grid/chat/members/maps/members.ts
```diff
@@ -42,13 +47,11 @@ export const getMembersTest = (
           return;
         }
 
+        input.page = cursor;
```
freaz, line 50: if cursor exists we will set it as `page` in input.

Jakub-Vacek (approved):
Looks ok 👍

PR #363, opened 2023-02-24
feat:add list candidates for greenhouse

cigar-galaxy82:
## Description

Add list candidates for greenhouse

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] I haven't repeated the code. (DRY)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #365, opened 2023-02-27
fix: run CI/CD pipeline for public fork pull requests

tmladek:
## Description

Hopefully also runs CI/CD pipeline for outside collaborators' pull requests (e.g. #363)
## Motivation and Context

Going forward, we may expect more PRs that do not originate from the origin `superfaceai/station` repo.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1,10 +1,9 @@
 name: CI / CD
 on:
-  push:
   pull_request:
-    types: [opened, reopened]
-concurrency:
-  group: ${{ github.ref }}
+  push:
+    branches:
+      - main
 jobs:
   lint_tools:
     name: Tools Lint and Format
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -3,6 +3,8 @@ on:
   push:
   pull_request:
     types: [opened, reopened]
+concurrency:
+  group: ${{ github.ref }}
 jobs:
   lint_tools:
     name: Tools Lint and Format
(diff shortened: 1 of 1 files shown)


freaz (dismissed):
Can't tell if it is 100% right, but seems legit.

jnv:
@tmladek Consider changing the triggers to run action only on push to main. That's actually the recommended approach: https://github.com/orgs/community/discussions/26276#discussioncomment-3251140

.github/workflows/ci_cd.yml
```diff
@@ -1,5 +1,10 @@
 name: CI / CD
-on: push
+on:
+  push:
+  pull_request:
+    types: [opened, reopened]
+concurrency:
+  group: ${{ github.ref }}
```
jnv, line 7: issue(blocking): Per discussion with @freaz we concluded that this isn't correct approach. As far as we understand it, this will run workflow on PR only when it's opened, but not on consecutive pushes from forked repository.
Here's a suggested fix (pending internal announcement):

```suggestion
on:
  pull_request:
  push:
    branches:
      - main
```

jnv (changes requested):

.github/workflows/ci_cd.yml
tmladek: Just to make sure, does this trigger when a PR is "updated" by pushing new commits to the source branch?

I don't think that's technically an action *on the pull request*, and [there's no activity type](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) that obviously stands out to me as indicating this happened; which would make sense, as that has nothing to do with the PR itself, because it's a *push* to the source branch. But that would get filtered out, because it's not a push to `main`.

Am I making sense? What happens in that case?

jnv (approved):

PR #369, opened 2023-03-07
fix(deepl): Send payload in body, parametrize provider

jnv:
**This PR changes the `deepl` provider!**

## Description

Current DeepL map for `language/translate` is broken, basic request returns 400. It looks like DeepL stopped accepting payload in query parameters, and requires data to be passed in body.

Furthermore, it's not possible to use DeepL API Pro in current map, so I've added new `SUBDOMAIN` parameter. It defaults to `api-free` for the free API.

One caveat though: DeepL defines the following [authentication schema](https://www.deepl.com/docs-api/api-access/authentication/) in HTTP header:

```
Authorization: DeepL-Auth-Key [yourAuthKey]
```

However, we don't have any schema which allows formatting the Authorization header. Luckily the currently used scheme with `auth_key` query parameter works, but it's not documented and probably shouldn't be used.

## Motivation and Context

Reported by user on Discord.

## Types of changes

## Checklist:

### DIFF
--- a/superface/super.json
+++ b/superface/super.json
@@ -1331,7 +1331,10 @@
           "id": "apikey",
           "apikey": "$DEEPL_API_KEY"
         }
-      ]
+      ],
+      "parameters": {
+        "SUBDOMAIN": "api-free"
+      }
     },
     "overpass-de": {
       "file": "../providers/overpass-de.json",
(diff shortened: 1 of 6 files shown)


superface/super.json
```diff
@@ -1331,7 +1331,10 @@
           "id": "apikey",
           "apikey": "$DEEPL_API_KEY"
         }
-      ]
+      ],
+      "parameters": {
+        "SUBDOMAIN": "api-free"
```
freaz, line 1336: Is this here to be explicit or the default in parameter definition doesn't work?

freaz (approved):
Last 6 months `language/translate` used mainly with mock and 7 times failed with DeepL. So updating provider should be safe.

image

PR #370, opened 2023-03-13
Feature/recruitment create lead greenhouse

cigar-galaxy82:
## Types of changes

## Checklist:

- [] I haven't repeated the code. (DRY)
- [] All new and existing tests passed.

### DIFF
--- a/grid/recruitment/create-lead/maps/greenhouse.suma
+++ b/grid/recruitment/create-lead/maps/greenhouse.suma
@@ -2,7 +2,6 @@ profile = "recruitment/create-lead@1.0"
 provider = "greenhouse"
 
 map CreateLead {
-  
 
   call CheckJobExists (
     jobId = input.jobId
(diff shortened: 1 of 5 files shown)


grid/recruitment/create-lead/maps/greenhouse.suma
```diff
@@ -0,0 +1,98 @@
...
+      map error {
+        title = 'Unauthenticated'
+        detail = body.message
+        code = 'Unauthenticated'
+      }
+    }
+
+    response 422 "application/json" {
+      map error {
+        title = 'Unprocessable Entity'
```
tmladek, line 73: Did you encounter this error in practice? When?

("Bad Request" ideally should never happen, as all requests made from the map should be valid - and a proper error should be thrown instead)

tmladek (changes requested):
Looks good, thanks! Just one question:

grid/recruitment/create-lead/maps/greenhouse.suma
cigar-galaxy82: I was able to recreate this while testing the API with postman but I tried out with the maps and it's not possible I remove this and push the changes.

tmladek (approved):
Thanks! Looks good to me :)

PR #373, opened 2023-03-28
fix(social-media/publish-post): Allow null fields

jnv:
## Description

This changes `string!` input fields to `string` to allow null values for all parameters.

Bumped patch version since this is fully backward compatible change.

## Motivation and Context

OneService interprets `field!` as required, which means we need to pass an empty string even for optional values.

Other social-media profiles don't use this type, so this is aligning the interface.

## Types of changes

## Checklist:


janhalama (approved):
LGTM, makes sense to align social media profile interfaces

PR #346, opened 2023-01-26
fix: ATS tests

tmladek:
## Description

Fixes `create-candidate` tests:
- use `expect(() => result.unwrap()).toThrow()` instead of just testing for `isOk()` to provide better errors
- actually check for error codes of expected errors

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I haven't repeated the code. (DRY)
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #379, opened 2023-04-18
fix: ip geolocation tests check result status

tmladek:
## Description

Make test for `address/ip-geolocation` also test result status instead of just whether they match the snapshot.

## Motivation and Context

Without the check, an erroneous snapshot may be recorded, while the tests are still passing.

(I think this should also be added to `CONVENTIONS`?)

## Types of changes

## Checklist:

### DIFF
--- a/grid/address/ip-geolocation/maps/ip-geocoding.ts
+++ b/grid/address/ip-geolocation/maps/ip-geocoding.ts
@@ -20,39 +20,39 @@ export const ipGeolocationTest = (
     describe('IpGeolocation', () => {
       describe('when api key is valid', () => {
         it('should return geolocation coordinates and address for valid IP address', async () => {
-          await expect(
-            superfaceTest.run(
-              {
-                useCase: 'IpGeolocation',
-                input: { ipAddress: '8.8.8.8' },
-              },
-              recordingOptions
-            )
-          ).resolves.toMatchSnapshot();
+          const result = await superfaceTest.run(
+            {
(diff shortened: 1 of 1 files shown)


jnv (changes requested):
suggestion: I agree in priciple, but consider using `result.isErr()` or `result.isOk()` instead of catching an error:

```
expect(result.isErr()).toBe(false)
```

As for adding this to conventions, I think a preferred way would be to adjust the testing library to explicitly opt-in to snapshotting the errors or providing a convenience function for that. This was discussed and agreed upon internally, but wasn't prioritized, unfortunately.

tmladek:
That's actually precisely the pattern I'm trying to avoid; simply checking for `isOk()` or `isErr()` doesn't provide as much information in case an error occurs, since all you get is that `isOk() is false`. On the other hand, testing `unwrap()` makes the tests output the actual error in case of failure, meaning you know right away where's the issue (or where to start looking, anyway).

>I think a preferred way would be to adjust the testing library to explicitly opt-in to snapshotting the errors or providing a convenience function for that.

But I think you're addressing that here, right? As you say though...

>This was discussed and agreed upon internally, but wasn't prioritized, unfortunately.

Hence why I've opted to write tests this way for now, for example in #346 too

jnv (approved):
Thanks for the explanation. I agree so let's do it this way. Hopefully we'll get chance to improve this pattern with the upcoming SDK changes.

PR #384, opened 2023-05-05
[WIP]feat:  shopify demo use cases

Jakub-Vacek:
## Description

This PR adds basic use cases for Shopify. Needs to be retested and polished before merging.

## Types of changes

## Checklist:

### DIFF
--- a/grid/product-management/product-update/maps/shopify.suma.js
+++ b/grid/product-management/product-update/maps/shopify.suma.js
@@ -1,0 +1,35 @@
+function UpdateProduct({ input, parameters, services }) {
+  const updateProductResult = updateProduct(input, services, parameters);
+
+  return updateProductResult;
+}
+
+function updateProduct(input, services, parameters) {
+  const url = `${services.default}/admin/api/${parameters.api_version}/products/${input.product.id}.json`;
+  const options = {
+    method: 'PUT',
+    body: {
+      product: input.product,
+    },
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    security: 'apiKey',
+  };
+
+  const response = std.unstable.fetch(url, options).response();
+  const body = response.bodyAuto() ?? {};
+
+  if (response.status !== 200) {
+    throw new std.unstable.MapError({
+      code: response.status,
+      message: 'HTTP call failed',
+    });
+  }
+
+  const result = {
+    product: body.product,
+  };
+
+  return result;
+}
--- a/grid/webhook-management/subscribe-webhook/map/shopify.suma.js
+++ b/grid/webhook-management/subscribe-webhook/map/shopify.suma.js
@@ -1,0 +1,44 @@
+function WebhookSubscription({ input, parameters, services }) {
+  const createWebhookResult = createWebhook(input, services, parameters);
+
+  return createWebhookResult;
+}
+
+function createWebhook(input, services, parameters) {
+  const url = `${services.default}admin/api/${parameters.api_version}/webhooks.json`;
+  const options = {
+    method: 'POST',
+    body: input.webhook,
+    headers: {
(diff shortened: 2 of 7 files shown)


providers/shopify.json
```diff
@@ -0,0 +1,29 @@
...
+  ],
+  "parameters": [
+    {
+      "name": "SHOP",
+      "description": "Your shop. You can find it in Shopify account settings."
+    },
+    {
+      "name": "api_version",
+      "description": "The version of the API being used.",
+      "default": "2023-04"
```
TheEdward162, line 26: question: I feel like this should be a part of the map, no? As in, the map isn't expected to work with _any_ API version, just the one it was written against, so I don't really understand the point of this being a parameter.

grid/customer-management/create-customer/maps/shopify.suma.js
```diff
@@ -0,0 +1,65 @@
+function CreateCustomer({ input, parameters, services }) {
```
TheEdward162, line 1: nitpick: Ideally this function is the entry - there is no need for indirection to call `createCustomer` since its body can just be directly in this function.

grid/product-management/product-update/maps/shopify.suma.js
```diff
@@ -0,0 +1,40 @@
...
+
+  const response = std.unstable.fetch(url, options).response();
+  const body = response.bodyAuto() ?? {};
+
+  if (response.status !== 200) {
+    throw new std.unstable.MapError({
+      errors: [
+        {
+          code: response.status.toString(),
+          field: 'HTTP',
```
TheEdward162, line 28: nitpick: not sure why this error has `field` - it just makes it different from the other profiles and doesn't really add much imo

grid/webhook-management/subscribe-webhook/map/shopify.suma.js
```diff
@@ -0,0 +1,49 @@
...
+  const response = std.unstable.fetch(url, options).response();
+  const body = response.bodyAuto() ?? {};
+
+  if (response.status !== 201) {
+    throw new std.unstable.MapError({
+      errors: [
+        {
+          code: response.status.toString(),
+          message: 'Failed to create webhook',
+          options: body,
```
TheEdward162, line 27: nitpick: I think this is wrong - body is not necessarily an object as profile requires. Also I think the name of the field is unclear.

Jakub-Vacek: Yes, I missed this.

grid/customer-management/create-customer/maps/shopify.suma.js
Jakub-Vacek: Yep, this is on the todo list :)

zdne (commented):
@Jakub-Vacek seems like the results miss description fields why inputs have them?

zdne:
@Jakub-Vacek also the "error" handling seems a bit naive. For instance, rate limits are problems with Shopify API and could be good to handle them well.

Jakub-Vacek:
> @Jakub-Vacek seems like the results miss description fields why inputs have them?

I prompted AI to do it this way - input is provided by user sou some guidance can be useful. We can remove them from input or add them to result as wall. 



> @Jakub-Vacek also the "error" handling seems a bit naive. For instance, rate limits are problems with Shopify API and could be good to handle them well.

It is naive, AI lacks info about rate limiting (I am trying to add it now) so there is room for improvement.

zdne:
> > @Jakub-Vacek seems like the results miss description fields why inputs have them?
> 
> I prompted AI to do it this way - input is provided by user sou some guidance can be useful. We can remove them from input or add them to result as wall.

I like it and, if feasible, I would add it to the result too!

PR #386, opened 2023-06-06
Add `captcha/verify-token` profile with `mock` and `cloudflare-turnstile` maps

janhalama:
## Description

This PR adds `captcha/verify-token` profile with `mock` and `cloudflare-turnstile` maps.

## Motivation and Context

`captcha/verify-token` profile is referenced in a new blog post comparing CAPTCHA providers. To keep the PR small, only the Cloudflare Turnstile provider map is included. Google reCAPTCHA and hCaptcha providers will come later in separate PRs.

## Types of changes

## Checklist:


providers/cloudflare.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "cloudflare",
+  "services": [
+    {
+      "id": "challenges",
+      "baseUrl": "https://challenges.cloudflare.com/"
+    }
+  ],
+  "defaultService": "challenges",
```
jnv, line 9: question: Is it necessary to define default service? Assuming we have Cloudflare for all of their APIs, I find it unlikely most maps will use the challenges service.

providers/cloudflare.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "cloudflare",
```
jnv, line 2: question: Have you considered providing a service-specific provider definition, e.g. `cloudflare-turnstile`?

jnv (changes requested):
In principle OK, but I have issue about how error conditions and challenge failure are communicated.

Also, consider specializing the provider (`cloudflare-turnstile` instead of `cloudflare`)

providers/cloudflare.json
janhalama: Good point. The default service should not be Turnstile challenges service address.

providers/cloudflare.json
janhalama: Yes, I have considered this option, but since OneSDK supports selecting the service in the map, I believe that having a `cloudflare` provider is more appropriate. Do you see any drawbacks to this solution?

jnv: For one thing, it's recommended by conventions (see google-apis-*).

The biggest drawback is that you haven't covered all Cloudflare APIs and can't cover the future ones. And you know how problematic it is to change published providers.
However, I can't say if that's still a relevant issue given the new OneSDK.

janhalama: Thanks for explaining it. I changed the provider name to `cloudflare-turnstile` which also solves the problem with default service in previous comment.

grid/captcha/verify-token/profile.supr
```diff
@@ -0,0 +1,67 @@
...
+    "Timestamp of the challenge."
+    challengeTimeStamp! string!
+  
+    "Hostname of the site where the CAPTCHA was solved."
+    hostname! string!
+  }
+
+  error {
+    "List of error codes related to the failed token validation. Refer to your CAPTCHA provider error code list. In case of invalid provider response `invalid-provider-response` code is returned."
+    errorCodes! [string!]!
```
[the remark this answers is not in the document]
janhalama: I went through the list of error codes for all 3 providers and added all of them to the profile. The list is pretty normalised across all providers.

> Also, should failing CAPTCHA validation be an error condition or a regular response?
IMO failing validation should return error, that is how I designed the profile.

jnv (approved):
Thanks for the changes, I have two small non-blocking notes. I think the communication around error could be still improved, but let's leave it for now; enums are already good start.

